PR #229, opened 2018-06-14
Scroll indicators update - positioning correction, support for both indicators at the same time

janek:
Corrects positioning, adds configurable indicator insets, adds conditional positioning when one/both indicators are visible.

**Type of change:** Bug fix + feature

## Todos
* [x] corrected positioning
* [x] added support for both indicators at the same time 
* [x] implemented default/base apple inset values
* [x] added support for configurable `scrollIndicatorInsets` (as per apple docs)
* [x] fixed other subviews of UIScrollView being added in front of indicators
 
## Expected behavior
Scroll indicators should now position and style themselves in a way indistinguishable from Apple's. 

#### Testing Details

Tested visually on NativePlayer with some additional marker views for better confidence.
Tested with a smaller dedicated app, both visually and numerically.

### Screenshots 
Current (desired) positioning of indicators:

screenshot 2019-03-05 at 13 58 14

### Please check if the PR fulfills these requirements

* [x] Self-review: I am confident this is the simplest and clearest way to achieve the expected behaviour
* [x] There are no dependencies on other PRs or I have linked dependencies through Zenhub
* [ ] The commit messages are clean and understandable
* [ ] Tests for the changes have been added (for bug fixes / features)
* [ ] Code runs on all relevant platforms (if major change: screenshots attached)


ephemer (changes requested):

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -12,26 +12,55 @@
 internal extension UIScrollView {
     internal static let indicatorDistanceFromScrollViewFrame: CGFloat = 2.5
 
-    internal func indicatorOffsetsInContentSpace() -> (horizontal: CGFloat, vertical: CGFloat) {
-        let totalContentArea = (
-            horizontal: contentInset.left + contentSize.width + contentInset.right,
-            vertical: contentInset.top + contentSize.height + contentInset.bottom
-        )
+    internal var indicatorLengths: (horizontal: CGFloat, vertical: CGFloat) {
```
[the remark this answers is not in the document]
janekszynal: clear about get{}, just slipped my mind that it's legal without it - correcting in the next commit.

controlled access levels and I was able to change a few to private, leaving only three functions that are indeed called from outside the file (from `UIScrollView.swift` in case of all 3)

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -12,26 +12,55 @@
...
+        }
+    }
 
-        let scrollViewProgress = (
-            horizontal: (contentInset.left + contentOffset.x) / (totalContentArea.horizontal + bounds.width),
-            vertical: (contentInset.top + contentOffset.y) / (totalContentArea.vertical + bounds.height)
-        )
+    internal var indicatorOffsetsInContentSpace: (horizontal: CGFloat, vertical: CGFloat) {
+        get {
+            let indicatorDistanceFromScrollViewFrame =  UIScrollView.indicatorDistanceFromScrollViewFrame
```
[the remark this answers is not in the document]
janekszynal: there was a conscious effort there (to have all numerical constants at the top of the file) - happy to discuss sometime, but for now just changing to what you suggested

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -12,26 +12,55 @@
...
-        return (
-            horizontal: contentOffset.x + indicatorOffsetInBounds.horizontal,
-            vertical: contentOffset.y + indicatorOffsetInBounds.vertical
-        )
+            let scrollViewProgress = (
+                horizontal: (contentInset.left + contentOffset.x) / (totalContentArea.horizontal - bounds.width),
+                vertical: (contentInset.top + contentOffset.y) / (totalContentArea.vertical - bounds.height)
+            )
+
+            let bothIndicatorsShowing = shouldLayoutHorizontalScrollIndicator && shouldLayoutVerticalScrollIndicator
```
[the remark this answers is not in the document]
janekszynal: renamed.
think I assigned the right meaning to how I was thinking about that constant, but maybe had a wrong picture of that

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -12,26 +12,55 @@
...
+
+            let additionalSpacingToPreventOverlap = (
+                horizontal: bothIndicatorsShowing ? 2*indicatorDistanceFromScrollViewFrame : 0,
```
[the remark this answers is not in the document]
janekszynal: done

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -42,28 +71,27 @@ internal extension UIScrollView {
...
+            horizontal: indicatorThickness + scrollIndicatorInsets.bottom,
+            vertical: indicatorThickness + scrollIndicatorInsets.right
+        )
 
         if shouldLayoutHorizontalScrollIndicator {
             horizontalScrollIndicator.frame = CGRect(
-                x: indicatorDistanceFromScrollViewFrame + indicatorOffsets.horizontal,
-                y: bounds.height - (indicatorThickness + indicatorDistanceFromScrollViewFrame),
+                x: scrollIndicatorInsets.left + indicatorOffsetsInContentSpace.horizontal, //scrolling direction
+                y: bounds.size.height + contentOffset.y - distanceFromFrame.horizontal,
```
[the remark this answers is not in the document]
janekszynal: done

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -42,28 +71,27 @@ internal extension UIScrollView {
...
+        let distanceFromFrame = (
+            horizontal: indicatorThickness + scrollIndicatorInsets.bottom,
+            vertical: indicatorThickness + scrollIndicatorInsets.right
+        )
 
         if shouldLayoutHorizontalScrollIndicator {
             horizontalScrollIndicator.frame = CGRect(
-                x: indicatorDistanceFromScrollViewFrame + indicatorOffsets.horizontal,
-                y: bounds.height - (indicatorThickness + indicatorDistanceFromScrollViewFrame),
+                x: scrollIndicatorInsets.left + indicatorOffsetsInContentSpace.horizontal, //scrolling direction
```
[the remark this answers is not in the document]
janekszynal: this was just for me and should have been removed before pushing. sorry, fixed.

Sources/UIScrollView.swift
```diff
@@ -120,10 +130,29 @@ open class UIScrollView: UIView {
...
+    public var indicatorStyle: UIScrollViewIndicatorStyle = .`default` {
+        didSet { applyScrollIndicatorsStyle() }
+    }
+    
+    private var customScrollIndicatorInsets = UIEdgeInsets.zero
+    private let baseScrollIndicatorInsets = UIEdgeInsets(top: 2.5, left: 2.5, bottom: 2.5, right: 2.5)
+    
+    public var scrollIndicatorInsets: UIEdgeInsets {
+        get {
+            let totalScrollIndicatorInsets = UIEdgeInsets(top: baseScrollIndicatorInsets.top + customScrollIndicatorInsets.top,
```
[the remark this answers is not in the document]
janekszynal: you're right, this needs correction, I'm working on it now. 

They are indeed 0,0,0,0 when logged. 
I took my values from observing the placement, they are indeed inset 2.5 from the sides - but when it's set to 0. It's a bit confusing, but I'll come up with something.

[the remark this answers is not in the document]
janekszynal: ok, done, this actually made code clearer aswell. thanks for noticing it!

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -42,28 +67,27 @@ internal extension UIScrollView {
...
 
-
     internal func layoutScrollIndicatorsIfNeeded() {
         guard shouldLayoutHorizontalScrollIndicator || shouldLayoutVerticalScrollIndicator else { return }
-
-        let indicatorDistanceFromScrollViewFrame = UIScrollView.indicatorDistanceFromScrollViewFrame
-        let indicatorOffsets = indicatorOffsetsInContentSpace()
-        let indicatorLengths = (horizontal: (bounds.width / contentSize.width) * bounds.width,
-                                vertical: (bounds.height / contentSize.height) * bounds.height)
+
```
ephemer: whitespace

rikner:
@ephemer @janekszynal is there anything preventing us from merging this? do we need any tests?

Sources/UIScrollView.swift
```diff
@@ -120,10 +127,24 @@ open class UIScrollView: UIView {
     open var contentInset: UIEdgeInsets = .zero
     open var contentSize: CGSize = .zero
 
-
     // MARK: Scroll Indicators
 
     let indicatorThickness: CGFloat = 2.5
+    
+    public var indicatorStyle: UIScrollViewIndicatorStyle = .`default` {
```
rikner: `UIScrollViewIndicatorStyle` is now `UIScrollView.IndicatorStyle`: https://developer.apple.com/documentation/uikit/uiscrollview/indicatorstyle

rikner (commented):
@janek maybe you can add a screenshot of your testapp, which shows both indicators. apart from that and the little API change we need to conform to I think this is ready to get merged soon.

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -42,28 +67,27 @@ internal extension UIScrollView {
         return showsVerticalScrollIndicator && contentSize.height > bounds.height
     }
 
-
     internal func layoutScrollIndicatorsIfNeeded() {
         guard shouldLayoutHorizontalScrollIndicator || shouldLayoutVerticalScrollIndicator else { return }
```
ephemer: iOS always lays out the scroll indicators I think, and there was an advantage to this -> I seem to remember finding a bug where the indicators appeared in the wrong spot after being hidden. I think we need to revisit this

Sources/UIScrollView+indicatorsInternal.swift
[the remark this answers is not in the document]
janek: Access levels were still not corrected, must have not pushed some changes. pushed them now.

Sources/UIScrollView.swift
janek: done

Sources/UIScrollView.swift
```diff
@@ -120,10 +127,24 @@ open class UIScrollView: UIView {
...
+        didSet { applyScrollIndicatorsStyle() }
+    }
+    
+    private let baseScrollIndicatorInsets = UIEdgeInsets(top: 2.5, left: 2.5, bottom: 2.5, right: 2.5)
```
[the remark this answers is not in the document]
janek: done

Sources/UIScrollView+indicatorsInternal.swift
janek: hey, I ran a test app with a scrollview on pure iOS, with showsXScrollIndicator off for both dimensions, and then on. I then used `po scrollView.subviews` in the debugger. If showsXScrollIndicator is off, they indicators are not in the view hierarchy at all, so it seems like iOS does not position them in that case. I could imagine that iOS might save positions for indicators without adding them to view hierarchy just yet, so maybe that's what happens - but no idea how to test that.

In light of this, what do we want to do? 
From the perspective of code changes, laying them out in all cases should be very easy and help us get rid of a few lines of code. So I absolutely don't mind doing it if you feel that's the way to go.

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,22 @@ open class UIScrollView: UIView {
...
+    open override func addSubview(_ view: UIView) {
+        if view == horizontalScrollIndicator || view == verticalScrollIndicator {
+            super.addSubview(view)
+        } else {
+            super.insertSubview(view, at: subviews.count - 2)
+        }
+    }
+    
+    open override func insertSubview(_ view: UIView, at index: Int) {
+        if index > subviews.count - 2 {
```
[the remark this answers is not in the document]
janek: checked on it now and it's changed to your suggestion. resolving.

Sources/UIScrollView+indicatorsInternal.swift
janek: seems to be corrected/not exisiting. resolving.

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,19 @@ open class UIScrollView: UIView {
             addSubview($0)
         }
     }
+    
+    // When adding subviews, make sure that the scroll indicators stay on top
+    open override func addSubview(_ view: UIView) {
+        if view == horizontalScrollIndicator || view == verticalScrollIndicator {
+            super.addSubview(view)
+        } else {
+            super.insertSubview(view, at: subviews.count - 2)
```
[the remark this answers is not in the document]
janek: I understand that this can be concerning. Also don't have anything breakthrough to propose, some small ideas: 
- make indicators their own class so it's more elegant (and safe?)  to check whether something is an indicator or not. 
- insted of inserting at end-2, we could check every element, starting from the last one, for being a scrollIndicator. if it is, we move one back. when it finally isn't, we insert. this should be a little safer.

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -34,44 +59,43 @@ internal extension UIScrollView {
...
     }
 
-    internal var shouldLayoutVerticalScrollIndicator: Bool {
+    private var shouldLayoutVerticalScrollIndicator: Bool {
         return showsVerticalScrollIndicator && contentSize.height > bounds.height
     }
 
-
-    internal func layoutScrollIndicatorsIfNeeded() {
+    func layoutScrollIndicatorsIfNeeded() {
```
ephemer: We have a convention of always writing `internal` to make it clear that that’s our intention. It’s the default in Swift but since we’re making a public-facing library our default is public/open.

Sources/UIScrollView+indicatorsInternal.swift
ephemer: Hi Janek, you can get a reference to the scroll indicator subviews (just iterate through them on an empty scroll view and find the UIImageViews)  and then hide them. You should be able to print their positions at every pixel scrolled. Would be cool if you could check this. I think it makes sense to remove them from the hierarchy of they’re hidden, but laying them out shouldn’t be a big performance issue (if that’s what iOS does)

Sources/UIScrollView.swift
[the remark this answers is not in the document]
ephemer: I think we need to avoid hard coding 2 because if as you say the subviews are removed on iOS when hidden we might as well do that

What about the 
insertSubview(below: otherSubview) API?

Sources/UIScrollView+indicatorsInternal.swift
janek: Yeah, I was wondering why the `internals` are explicitly there. Guess I was on the wrong side of https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence :d 

brought them back now

Sources/UIScrollView+indicatorsInternal.swift
janek: yes, I can get the reference and print positions, even if I make them invisible with `isHidden`.

but if I set `showsVerticalScrollIndicator` for `false`, that indicator will no longer be in the view hierarchy and therefore probably impossible to get a reference to.

I checked back with our code and this is where we differ - we always have indicators in the hierarchy and just make them invisible if the UIKit user wants to hide them.

So if they are set to hidden by UIKit API, we keep them in the hierarchy, but don't update their positions, and then start updating it when they become visible. We could:
1. keep it as it is 
2. always update their positions, even when theyr are invisible
3. try to get as close to iOS as possible and actually remove them and add back as needed.

1. and 2. are equally easy. We are in state 1 and if we want 2, we should just merge this new small PR I opened (#284). It's set to merge into this branch, not master.

3. is more work and I worry it might also complicate things with our inserting/removing subviews code, so I'd rather not do it now - but again, it seems the closest to iOS. Maybe let's go with 1 or 2 and create a low-prio issue for 3?

janek:
hey @ephemer, I can't comment on your last comment, not sure why. It might be because I changed the code it was referencing (and it got auto-marked as outdated) or because it was a comment to the PR in general and not a line in code (and therefore doesn't have a reply thread).

Anyway, I changed the add/insert subview code. Couldn't avoid writing a few extra lines, but I think it's all pretty understandable and therefore no big harm done. Let me know what you think.

Sources/UIScrollView+indicatorsInternal.swift
janek: update: I now changed the insert/add subview code - didn't take long at all and should now be robust w.r.t indicators sometimes not being there. It therefore also shouldn't be too much pain to implement v3. I'll still wait for your opinion before starting on that.

ephemer:
Hi @janek, I'd really like to get this PR merged because it's been open for a long time. My main concern is still this ordering of the indicators. I have the feeling that part could be buggy still. Let's take some time when you're back in the office to go through this and get it merged

janek:
yeah, I would love that too, it's been very long. Let's try tomorrow!

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,50 @@ open class UIScrollView: UIView {
             addSubview($0)
         }
     }
+    
+    // When adding subviews, make sure that the scroll indicators stay on top.
+    // Subview list goes back-to-front, so indicators need to take the last two positions.
+    // We use the earliestIndicatorInSubviewHierarchy variable to make sure new views are always inserted below them.
+    private lazy var earliestIndicatorInSubviewHierarchy: UIView? = {
+        let indexOfHorizontal = subviews.index(of: horizontalScrollIndicator)
```
ephemer: this entire function can be written much more simply:

`return subviews.first(where: { $0 === horizontalScrollIndicator || $0 === verticalScrollIndicator })`

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,50 @@ open class UIScrollView: UIView {
...
+        }
+    }
+
+
+    open override func insertSubview(_ view: UIView, at index: Int) {
+        var indexOfLastIndicator = Int.max
+        if let earliestIndicatorInSubviewHierarchy = earliestIndicatorInSubviewHierarchy {
+            indexOfLastIndicator = subviews.index(of: earliestIndicatorInSubviewHierarchy)!
+        }
+        super.insertSubview(view, at: min(index, indexOfLastIndicator-1))
```
ephemer: linting: `indexOfLastIndicator - 1`

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,50 @@ open class UIScrollView: UIView {
...
+        } else {
+            return verticalScrollIndicator
+        }
+    }()
+
+
+    open override func addSubview(_ view: UIView) {
+        // If the view we're adding is an indicator, just add it.
+        // If not, check for indicators and insert below the last one present.
+        if view == horizontalScrollIndicator || view == verticalScrollIndicator {
```
ephemer: this can be simplified too:

```
if view !== horizontalScrollIndicator, view !== verticalScrollIndicator, let earliestIndicatorInSubviewHierarchy = earliestIndicatorInSubviewHierarchy  {
    super.insertSubview(view, belowSubview: earliestIndicatorInSubviewHierarchy)
} else {
    super.addSubview(view)
}
```

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,50 @@ open class UIScrollView: UIView {
...
+            super.addSubview(view)
+        } else {
+            if let earliestIndicatorInSubviewHierarchy = earliestIndicatorInSubviewHierarchy {
```
ephemer: I'm not sure on the naming of this: `subviewHierarchy` is overly descriptive to me, and `earliest` is imprecise. What does earliest even mean in terms of array indexes or visual hierarchy? Is it `firstIndicatorInSubviews` maybe?

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,50 @@ open class UIScrollView: UIView {
...
+                super.insertSubview(view, belowSubview: earliestIndicatorInSubviewHierarchy)
+            } else {
+                super.addSubview(view)
+            }
+        }
+    }
+
+
+    open override func insertSubview(_ view: UIView, at index: Int) {
+        var indexOfLastIndicator = Int.max
```
ephemer: I'm not a big fan of this mutable style. I think it's harder to read as is, and prone to errors if we need to change it.

I think we can just have two code paths, one that does the `min` comparison, and one that just adds the view at the specified index.

Sources/UIScrollView.swift
```diff
@@ -120,11 +158,27 @@ open class UIScrollView: UIView {
...
     // MARK: Scroll Indicators
 
+    // Matching Apple's value
     let indicatorThickness: CGFloat = 2.5
 
+    // Determined experimentally to be as close as possible to Apple's (likely exactly matching)
+    private let baseScrollIndicatorInsets = UIEdgeInsets(top: 2.5, left: 2.5, bottom: 2.5, right: 2.5)
+
+    public var indicatorStyle: IndicatorStyle = .`default` {
+        didSet { applyScrollIndicatorsStyle() }
```
ephemer: just making sure `applyScrollIndicatorsStyle` is called elsewhere (as well) - `didSet`s don't get called on `init`. If this is the case we should add a comment to that effect here too

Sources/UIScrollView+indicatorsInternal.swift
```diff
@@ -6,26 +6,51 @@
...
         let scrollViewProgress = (
-            horizontal: (contentInset.left + contentOffset.x) / (totalContentArea.horizontal + bounds.width),
-            vertical: (contentInset.top + contentOffset.y) / (totalContentArea.vertical + bounds.height)
+            horizontal: (contentInset.left + contentOffset.x) / (totalContentArea.horizontal - bounds.width),
+            vertical: (contentInset.top + contentOffset.y) / (totalContentArea.vertical - bounds.height)
+        )
+
+        let shouldShowBothIndicators = shouldLayoutHorizontalScrollIndicator && shouldLayoutVerticalScrollIndicator
+
+        let additionalSpacingToPreventOverlap = (
```
ephemer: Did we make these distances up or are they iOS style? We should comment which

Sources/UIScrollView.swift
janek: _nice_

Sources/UIScrollView.swift
janek: done

Sources/UIScrollView.swift
janek: you're right, `firstIndicatorInSubviews` is way better (especially being consistent with the `.first(where:{}`) function). changed

Sources/UIScrollView.swift
janek: corrected

Sources/UIScrollView.swift
janek: done

Sources/UIScrollView.swift
```diff
@@ -77,6 +71,50 @@ open class UIScrollView: UIView {
...
+        // if both indicators are in the view hierarchy, return the one that is last
+        if indexOfHorizontal! < indexOfVertical! {
+            return horizontalScrollIndicator
+        } else {
+            return verticalScrollIndicator
+        }
+    }()
+
+
+    open override func addSubview(_ view: UIView) {
```
janek: I'm realizing that I should have overriden the `subviews` variable. Not only does it probably make `insertSubview` simpler, but should also just be there. Currently the behavior for `print(scrollView.subviews.count)` will differ from iOS in that it will also count the indicators.

Unfortunately, if I do override it, those versions of `addSubview` and `insertSubview` won't work as expected. I'm trying to figure it out now, noting here so all information about remaining todos is in one place.

Sources/UIScrollView+indicatorsInternal.swift
janek: done

Sources/UIScrollView.swift
janek: yes, it is being called on `init` separately. Not sure if I understand you correctly, should I add a comment stating that fact?

Sources/UIScrollView.swift
janek: Update: no need to override scrollviews, iOS does have them

Sources/UIScrollView+indicatorsInternal.swift
janek: update: merged #284

Sources/UIScrollView.swift
janek: consider:
```
super.addSubview(view)
addHorizontal
addVertical
```
for much simpler, a bit less performant code

janek: not sure what exactly I meant here, unfortunately - resolving thread

Sources/UIScrollView.swift
janek: I agree with this point and refactored - should be nice and readable now.

janek:
hey @ephemer :) I improved a lot here, addressing all the smaller comments you had. 

Unfortunately, I still have a concern:

if you check `scrollView.subviews` on iOS, it will not include scroll indicators. In our version, in the state from this PR, it will include them. We can address this here or with a new issue, depending on how demanding/important it is. I gave it a moment's thought and decided it's best to ask you for opinion before trying, and a tip on how to go about it.

ephemer:
@janek we could override `var subviews` to return a filtered version of `private var _subviews = [UIView]`. It's tricky though because AFAIK we're using `subviews` for some "internal" UIKit logic, which may or may not be correct.

But to me that's an issue that's totally separate from this PR. If you would like to update that let's do it for a practical (rather than idealistic) reason: right now I don't think it matters for us

janek:
cool, that's what I was thinking, too, for the implementation. Glad I had the right idea (and the right idea to ask). 

In that case, this is ready for re-review.

Sources/UIScrollView.swift
ephemer: @janek here you reported something different regarding subviews, or did I misunderstand one of the two pieces of info incorrectly :)

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,65 @@ class UIScrollViewTests: XCTestCase {
...
+        XCTAssertEqual(verticalIndicatorIndex, 4)
+        XCTAssertEqual(scrollView.subviews.count, 6)
+
+
+        let viewToBeInserted2 = UIView()
+        scrollView.insertSubview(viewToBeInserted2, at: 5) // position currently occupied by scroll indicator
+        let indexOfInsertedView2 = scrollView.subviews.index(of: viewToBeInserted1)!
+
+        // If we try to insert at a position occupied by or above indicators,
+        // the inserted view should 'slide down' and assume the highest position below indicators
```
ephemer: nice, these comments are helpful thanks!

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,65 @@ class UIScrollViewTests: XCTestCase {
...
+        XCTAssertEqual(horizontalIndicatorIndex, 3)
+        XCTAssertEqual(verticalIndicatorIndex, 4)
+        XCTAssertEqual(scrollView.subviews.count, 5)
+
+        let viewToBeInserted1 = UIView()
+        scrollView.insertSubview(viewToBeInserted1, at: 1)
+        let indexOfInsertedView1 = scrollView.subviews.index(of: viewToBeInserted1)!
+
+        // This might seem weird, but that's what happens in iOS:
+        // the inserted view and the one that was in its place before will have the same index
```
ephemer: is it possible this is only weird because of how the test is written?

It doesn't seem that weird to me that the indexes will overlap once we insert a new subview if we're not getting the _new_ indexes after inserting the subview?

What am I missing here? It seems like we should get the _new_ indexes here which would be less confusing?

samples/getting-started/DemoApp/ViewController.swift
```diff
@@ -15,12 +15,31 @@ class ViewController: UIViewController {
...
+
+        for (i, color) in colors.enumerated() {
+            let view = UIView(frame: CGRect(origin: CGPoint(x: 220+20*i, y: 10+20*i), size: CGSize(width: 100, height: 100)))
+            view.backgroundColor = color
+            views.append(view)
+            scrollView.addSubview(view)
+        }
+
+
+        let blueView = UIView(frame: CGRect(origin: CGPoint(x: 225+20*2 - 50, y: 15+20*2), size: CGSize(width: 150, height: 10)))
```
ephemer: please always use a space between binary operators
also, this line is very long, it would be great if you could split it up logically into

`let blueViewFrame = CGRect(x:,y:,width:,height:)` (avoiding the origin:,size: initializer)

also, these values seem pretty arbitrary, would be good if you could put them into named variables so they get some meaning

samples/getting-started/DemoApp/ViewController.swift
```diff
@@ -15,12 +15,31 @@ class ViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        label.text = "Hello World"
```
ephemer: please don't delete what we already had, but feel free to move the label somewhere else (outside the scroll view)

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,65 @@ class UIScrollViewTests: XCTestCase {
...
+        XCTAssertEqual(indexOfInsertedView1, 1)
+        XCTAssertEqual(label2Index, 1)
+        XCTAssertEqual(label3Index, 2)
+        XCTAssertEqual(horizontalIndicatorIndex, 3)
+        XCTAssertEqual(verticalIndicatorIndex, 4)
+        XCTAssertEqual(scrollView.subviews.count, 6)
+
+
+        let viewToBeInserted2 = UIView()
+        scrollView.insertSubview(viewToBeInserted2, at: 5) // position currently occupied by scroll indicator
```
ephemer: this test is too long. this could be two or three separate tests which would be much clearer if one of them fails

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,65 @@ class UIScrollViewTests: XCTestCase {
...
+        scrollView.insertSubview(viewToBeInserted1, at: 1)
+        let indexOfInsertedView1 = scrollView.subviews.index(of: viewToBeInserted1)!
+
+        // This might seem weird, but that's what happens in iOS:
+        // the inserted view and the one that was in its place before will have the same index
+        XCTAssertEqual(label1Index, 0)
+        XCTAssertEqual(indexOfInsertedView1, 1)
+        XCTAssertEqual(label2Index, 1)
+        XCTAssertEqual(label3Index, 2)
+        XCTAssertEqual(horizontalIndicatorIndex, 3)
```
ephemer: This is misleading at best: all of these indexes will be incorrect now, so what's the point in asserting what their _old_ value was? I'm really not happy with this, as written

UIKitTests/UIScrollViewTests.swift
ephemer: but this should just be its own standalone test. it tests one single behaviour that obviously must be true. if we insert a subview it should always be behind the indicators. end of test. if that breaks we know what to fix. does that make sense?

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,65 @@ class UIScrollViewTests: XCTestCase {
...
+        let label1Index = scrollView.subviews.index(of: view1)!
+        let label2Index = scrollView.subviews.index(of: view2)!
+        let label3Index = scrollView.subviews.index(of: view3)!
+
+        XCTAssert(horizontalIndicatorIndex > label1Index)
+        XCTAssert(verticalIndicatorIndex > label1Index)
+
+        XCTAssert(label2Index > label1Index)
+
+        XCTAssertEqual(label1Index, 0)
```
ephemer: what's the point in adding three subviews? we already have other tests that test the behaviour of `addSubview`. we're just writing more test code that we have to maintain.

how is this different to the case at the bottom?

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,65 @@ class UIScrollViewTests: XCTestCase {
...
+        scrollView.addSubview(view2)
+        scrollView.addSubview(view3)
+
+        let horizontalIndicatorIndex = scrollView.subviews.index(of: scrollView.horizontalScrollIndicator)!
+        let verticalIndicatorIndex = scrollView.subviews.index(of: scrollView.verticalScrollIndicator)!
+        let label1Index = scrollView.subviews.index(of: view1)!
+        let label2Index = scrollView.subviews.index(of: view2)!
+        let label3Index = scrollView.subviews.index(of: view3)!
+
+        XCTAssert(horizontalIndicatorIndex > label1Index)
```
ephemer: how is this different to the case at the bottom?

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,65 @@ class UIScrollViewTests: XCTestCase {
...
+        let view2 = UIView()
+        let view3 = UIView()
+
+        scrollView.addSubview(view1)
+        scrollView.addSubview(view2)
+        scrollView.addSubview(view3)
+
+        let horizontalIndicatorIndex = scrollView.subviews.index(of: scrollView.horizontalScrollIndicator)!
+        let verticalIndicatorIndex = scrollView.subviews.index(of: scrollView.verticalScrollIndicator)!
+        let label1Index = scrollView.subviews.index(of: view1)!
```
ephemer: I don't think it makes sense to put this into a variable, because (as mentioned below) it will be invalid as soon as we insert another view.

```
XCTAssertGreaterThan(
    scrollView.subviews.index(of: scrollView.horizontalScrollIndicator),
    scrollView.subviews.index(of: view1)
)
```

ephemer (changes requested):
Hi @janek, code-wise the library code seems fine but the tests are very shonky as is.

The tests contain false assumptions and hardcoded outdated values which don't mean anything. Sometimes we're calling an identical assertion twice (what's the point in calling `XCTAssertEqual` on two static values twice?), even though the second time the reality is that the underlying values have changed. It's not up to scratch as written.

Please try to rewrite the new tests in more finer-grained ones that test _one_ thing, that are clearly named so it's obvious what to fix if they break, that don't rely on hardcoded values if possible, and that test the absolute minimal amount possible while still ensuring behaviour. And please be careful to get the _new_ indexes after changing the order of an array.

samples/getting-started/DemoApp/ViewController.swift
janek: (pasting in both relevant places) 
Very sorry about this, the changes to this file were not meant to be submitted. They were useful to me while developing, but my goal wasn't to change the DemoApp. I was pushing it to keep the git index clean, with the intention of resetting the file before re-requesting review.

samples/getting-started/DemoApp/ViewController.swift
janek: (pasting in both relevant places) 
Very sorry about this, the changes to this file were not meant to be submitted. They were useful to me while developing, but my goal wasn't to change the DemoApp. I was pushing it to keep the git index clean, with the intention of resetting the file before re-requesting review.

ephemer: @janek I think it's great to add more functionality to the demo app, totally fine to leave it here, but please make the change additive rather than replacing what was there :)

samples/getting-started/DemoApp/ViewController.swift
```diff
@@ -15,12 +15,31 @@ class ViewController: UIViewController {
...
+            views.append(view)
+            scrollView.addSubview(view)
+        }
+
+
+        let blueView = UIView(frame: CGRect(origin: CGPoint(x: 225+20*2 - 50, y: 15+20*2), size: CGSize(width: 150, height: 10)))
+        blueView.backgroundColor = .blue
+
+        scrollView.insertSubview(blueView, at: 6)
+        print(scrollView.subviews.index(of: blueView) ?? -1)
```
ephemer: if you do decide to clean this up and keep the code, please remove any `print`s

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,56 @@ class UIScrollViewTests: XCTestCase {
...
+        )
+
+        XCTAssertGreaterThan(
+            scrollView.subviews.index(of: scrollView.verticalScrollIndicator)!,
+            scrollView.subviews.index(of: insertedMockView)!
+        )
+
+        // Currently broken
+        // XCTAssertEqual(indexOfInsertedView, mockViews.count)
+    }
```
janek: Very happy to update this when #309 is merged

UIKitTests/UIScrollViewTests.swift
janek: Thanks for this and all the other very valid concerns. I rewrote the test(s) in a way that hopefully takes into account all of them.

UIKitTests/UIScrollViewTests.swift
janek: (updated)

janek:
thanks @ephemer! I didn't reply to all of your comments, but I took them into account when rewriting tests. It's definitely a big improvement, and I'm quite happy. Also very happy to perfect them if there is anything else that you see could be improved :)

UIKitTests/UIScrollViewTests.swift
```diff
@@ -47,6 +47,42 @@ class UIScrollViewTests: XCTestCase {
...
+        // and we want newer 'normal' subviews to be above older ones
+
+        // XXX: iOS does not allow explicit access to scroll indicators via subviews.
+        // We should remove that, too - but for now it's the easy way to go and helpful in testing
+
+        let mockView = UIView()
+        scrollView.addSubview(mockView)
+
+        XCTAssertGreaterThan(
+            scrollView.subviews.index(of: scrollView.horizontalScrollIndicator)!,
```
ephemer: These tests look seriously amazing now. Great work @janek!

ephemer (approved):
Awesome! I am stoked with how great these new tests you wrote are looking now! Thanks for your work on this @janek

UIKitTests/UIScrollViewTests.swift
janek: that's too nice! thanks a lot!!

PR #265, opened 2018-10-11
Improve Touch Handling

### LINKED ISSUE #217: Android: Improve touch handling
SDL's touch handling makes it impossible to get accurate timestamps on events, which breaks our velocity scrolling and causes other weird behaviour.

Since we have a custom `custom/core/SDL_android.c`, we can override the entry point and do our own processing of the touch events, e.g. attaching correct timestamps, device/finger IDs and avoiding unnecessary unit conversions.

If we implement this ourselves in Swift, we will be part-way to being independent of SDL.

rikner:
Fixes #217 

**Type of change:** 

SDL's touch handling makes it impossible to get accurate timestamps on events, which breaks our velocity scrolling and causes other weird behaviour.


rikner:
@janekszynal you might want have a look at my changes.
I was able to correctly pass timestamp (maintaining the actual value) as `JavaLong` to the swift implementation of `onNativeTouch` by using `@_cdecl` instead of `@_silgen_name`.

src/main/java/org/libsdl/app/SDLOnTouchListener.kt
```diff
@@ -13,25 +13,14 @@ interface SDLOnTouchListener: View.OnTouchListener {
...
-
-        fun MotionEvent.touchValues(i: Int): TouchValues {
-            return TouchValues(
-                    getPointerId(i),
-                    getX(i) / mWidth,
-                    getY(i) / mHeight,
-                    // Pressure can be > 1.0 on some devices. See getPressure(i) docs.
-                    min(this.getPressure(i), 1.0f)
-            )
-        }
```
rikner: moved this to `private` top-level

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -14,6 +14,14 @@ extension UIApplication {
         var e = SDL_Event()
 
         while SDL_PollEvent(&e) == 1 {
+
+            #if !os(macOS)
```
michaelknoch:
```suggestion
            #if os(Android)
```

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -96,3 +104,130 @@ extension SDL_Event {
...
+                at: CGPoint(
+                    x: CGFloat(event.tfinger.x),
+                    y: CGFloat(event.tfinger.y)
+                ),
+                timestamp: event.timestampInSeconds
+            )
+
+            if
+                let firstExistingEvent = UIEvent.activeEvents.first,
+                let _ = firstExistingEvent.allTouches?.first(where: {$0.touchId == event.tfinger.fingerId} )
```
michaelknoch:
```suggestion
                let _ = firstExistingEvent.allTouches?.first(where: { $0.touchId == event.tfinger.fingerId })
```

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -96,3 +104,130 @@ extension SDL_Event {
...
+            {
+                //found a matching event, adding current touch to it and returning
```
michaelknoch:
```suggestion
                // Found a matching event, adding current touch to it and returning
```

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -96,3 +104,130 @@ extension SDL_Event {
...
+                firstExistingEvent.allTouches?.insert(newTouch)
+                return firstExistingEvent
+            } else {
+                //no matching event found, creating a new one
```
michaelknoch:
```suggestion
                // No matching event found, creating a new one
```

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -96,3 +104,130 @@ extension SDL_Event {
...
+                return UIEvent(touch: newTouch)
+            }
+
+        case SDL_FINGERMOTION:
+            if
+                let firstExistingEvent = UIEvent.activeEvents.first,
+                let matchingTouch = firstExistingEvent.allTouches?.first(where: { $0.touchId == event.tfinger.fingerId} )
```
michaelknoch:
```suggestion
                let matchingTouch = firstExistingEvent.allTouches?.first(where: { $0.touchId == event.tfinger.fingerId })
```

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -96,3 +104,130 @@ extension SDL_Event {
...
+                return firstExistingEvent
+            }
+            else {
+                return nil
+            }
+
+        case SDL_FINGERUP:
+            if
+                let firstExistingEvent = UIEvent.activeEvents.first,
+                let matchingTouch = firstExistingEvent.allTouches?.first(where: {$0.touchId == event.tfinger.fingerId} )
```
michaelknoch:
```suggestion
                let matchingTouch = firstExistingEvent.allTouches?.first(where: { $0.touchId == event.tfinger.fingerId })
```

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -96,3 +104,130 @@ extension SDL_Event {
...
+}
+
+extension UIEvent: CustomStringConvertible {
+    public var description: String {
+        var listOfTouches = ""
+        if let touches = allTouches {
+            for touch in touches {
+                listOfTouches += "\(touch.description),\n"
+            }
+        }
```
michaelknoch: we should implement this wit a map/join

Sources/UIScrollView+velocity.swift
```diff
@@ -20,7 +20,7 @@ extension UIScrollView {
         // Otherwise we could animate after scrolling quickly, pausing for a few seconds, then letting go
         let velocityIsLargeEnoughToDecelerate = (self.panGestureRecognizer.velocity(in: self).magnitude > 10)
 
-        let nonBoundsCheckedScrollAnimationDistance = self.weightedAverageVelocity * 0.74 // hand-tuned
+        let nonBoundsCheckedScrollAnimationDistance = self.weightedAverageVelocity * 0.4 // hand-tuned
```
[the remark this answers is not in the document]
rikner: I played with that value myself, and found that `0.4` seems reasonable, we still might want to fine tune it. Anyway I think it's weird that we had to change it in the first place. Not sure if this valueis just some random factor chosen by us or if it has any (physical) background (e.g. like the `animationTimeConstant` below)

ephemer:
Note to self: the current branch causes huge slowdowns with multitouch, we may need to reorganise some code to make this efficient

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -14,6 +14,14 @@ extension UIApplication {
         var e = SDL_Event()
 
         while SDL_PollEvent(&e) == 1 {
+
+            #if os(Android)
+            if let uiEvent = UIEvent.from(e) {
+                sendEvent(uiEvent)
+                break
+            }
+            #endif
```
michaelknoch: its not clear why we send send events for android only

Sources/UIEvent+fromSDL_Event.swift
```diff
@@ -0,0 +1,129 @@
+//
+//  UIEvent+fromSDL_Event.swift
+//  UIKit
+//
+//  Created by Geordie Jay on 05.11.18.
+//  Copyright © 2018 flowkey. All rights reserved.
+//
+
+#if os(Android)
```
michaelknoch: is this the current pattern for android only files? We could exclude them from the iOS and mac targets

Sources/UIEvent+fromSDL_Event.swift
```diff
@@ -0,0 +1,129 @@
...
+    env: UnsafeMutablePointer<JNIEnv>,
+    view: JavaObject,
+    touchDeviceID: JavaInt,
+    pointerFingerID: JavaInt,
+    action: JavaInt,
+    x: JavaFloat,
+    y: JavaFloat,
+    pressure: JavaFloat,
+    timestamp: JavaLong
+    ) {
```
michaelknoch:
```suggestion
) {
```
lint

Sources/UIApplication+handleSDLEvents.swift
ephemer: I'll write a comment, thanks

Sources/UIEvent+fromSDL_Event.swift
ephemer: Sounds good to me, thanks

Sources/UIEvent+fromSDL_Event.swift
ephemer: Thanks, a classic Xcode formatting fail..

rikner:
@ephemer @janek 

closing this PR in favor of 
https://github.com/flowkey/UIKit-cross-platform/pull/282/files
and
https://github.com/flowkey/UIKit-cross-platform/pull/283/files

PR #268, opened 2018-10-26
GLRenderer -> UIScreen

michaelknoch:
**Type of change:** 

tbd


Sources/Logging.swift
```diff
@@ -17,7 +17,7 @@ import Darwin.C.stdio
 private let loggingTag = "Swift"
 
 @discardableResult
-@_silgen_name("__android_log_write")
+@_cdecl("__android_log_write")
 public func android_log_write(_ prio: Int32, _ tag: UnsafePointer<CChar>, _ text: UnsafePointer<CChar>) -> Int32
```
michaelknoch: are you sure you want to use _cdecl here? i think _cdecl cannot annotate empty functions
https://github.com/apple/swift/blob/master/test/attr/attr_cdecl.swift#L9

Sources/SDL+JNIExtensions.swift
```diff
@@ -8,10 +8,10 @@
 
 import JNI
 
-@_silgen_name("Android_JNI_GetActivityClass")
+@_cdecl("Android_JNI_GetActivityClass")
 public func getSDLViewClass() -> JavaClass
 
-@_silgen_name("SDL_AndroidGetActivity")
+@_cdecl("SDL_AndroidGetActivity")
```
michaelknoch: same here

Sources/androidNativeInit.swift
```diff
@@ -10,13 +10,13 @@ import SDL
 import CJNI
 
 public struct UIKitAndroid {
-    @_silgen_name("SDL_Android_Init")
+    @_cdecl("SDL_Android_Init")
```
michaelknoch: for some reason _cdecl can only be used with top level functions

Sources/SDL+JNIExtensions.swift
ephemer: Hmm ok this actually makes sense because we’re not defining anything about the function with an empty declaration. It’s just signalling that the function exists.

Sources/androidNativeInit.swift
ephemer: This is more puzzling though. It shouldn’t make any difference whether it’s top-level or static from what I understand. I’d quite like to define these as cdecl because otherwise we may end up with unexpected behaviour again. So we should take them out of the struct and back to top level

michaelknoch:
Also tests are broken for a very interesting reason. (ButtonContentAlignmentTests.swift)

screenshot 2018-10-27 at 18 32 41

It makes sense that this is currently breaking here because UIApplication never inits the main window when running tests. It's very interesting that iOS seems to handle this case. Maybe UIScreen.main should always return a screen?

ephemer:
@michaelknoch I have made provisions for this in UIScreen.swift but not yet updated the tests.

It’s possible to init a dummy screen in the tests (there is an internal initialiser in that file that we can use with `@testable import`). I think this is better than the old solution

michaelknoch:
So we always have to manually init a UIScreen as soon as we are writing tests for UIView or something that depends on it? Sounds like an easy fix but also sounds kind of verbose and iOS seems to handle it different because the buttons tests are passing on iOS. @ephemer

ephemer:
@michaelknoch there are “global” setup / tear downs. I’d init the screen in there in the UIKit Test target. In practice it should be clean and also explicit, which is better than the kind of random implicit behaviour we used to have

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -6,6 +6,7 @@
 //  Copyright © 2018 flowkey. All rights reserved.
 //
 
+import Dispatch
```
ephemer: weg?

Sources/UIScreen.swift
```diff
@@ -6,29 +6,300 @@
...
     }
-}
 
-public extension UIScreen {
-    public static var main: UIScreen {
-        #if !DEBUG
-        // Crash in production when accessing this without a glRenderer.
-        if UIApplication.shared == nil {
-            preconditionFailure("Tried to get UIScreen.main dimensions before calling `UIApplicationMain`!")
+    // There is an inconsistency between Mac and Android when setting SDL_WINDOW_FULLSCREEN
```
ephemer: this comment is random

Sources/UIScreen.swift
```diff
@@ -6,29 +6,300 @@
...
+
+        // Fixes video surface visibility with transparent & opaque views in SDLSurface above
+        // by changing the alpha blend function to: src-alpha * (1 - dst-alpha) + dst-alpha
+        setShapeBlending(true)
+        setShapeBlendMode(GPU_BLEND_NORMAL_FACTOR_ALPHA)
+
+        clearErrors() // by now we have handled any errors we might have wanted to
+
+        UIFont.loadSystemFonts()
+        print("Inited UIScreen")
```
ephemer: remove

Sources/UIScreen.swift
```diff
@@ -6,29 +6,300 @@
...
+        guard let gpuContext = self.rawPointer.pointee.context else {
+            assertionFailure("glRenderer gpuContext not found")
+            return
+        }
+
+        let existingWindowID = gpuContext.pointee.windowID
+        let existingWindow = SDL_GetWindowFromID(existingWindowID)
+        SDL_DestroyWindow(existingWindow)
+        rawPointer = nil
+        print("Destroyed UIScreen")
```
ephemer: remove

Sources/UIScreen.swift
```diff
@@ -6,29 +6,300 @@
...
+    }
+
+    deinit {
+        destroy()
```
ephemer: do we use `destroy` in any other places any more? this was initially a workaround to allow us to deinit the Screen before the instance gets deinitted, but it may make sense to move the body of `destroy` into `deinit`

src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -177,7 +181,7 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
 
     override fun doFrame(frameTimeNanos: Long) {
         if (isRunning && mIsSurfaceReady) {
-            this.nativeRender()
+            this.nativeRenderAndProcessEvents()
```
ephemer: this should probably be called nativeProcessEventsAndRender (because it happens in that order)

Sources/UIApplicationMain.swift
```diff
@@ -55,13 +57,20 @@ internal func UIApplicationMain(
 }
 
 extension UIApplication {
-    static func restart() {
+    static func restart(_ onRestarted: (() -> Void)? = nil) {
```
ephemer: it's possible we're not using `onRestarted`, let's remove it if not

Sources/UIApplication.swift
```diff
@@ -53,17 +53,12 @@ open class UIApplication {
...
 
     public required init() {
-        UIFont.loadSystemFonts()
+        UIScreen.main = UIScreen()
     }
 
-    internal var glRenderer = GLRenderer()
-
     deinit {
         DisplayLink.activeDisplayLinks.removeAll()
```
ephemer: hmm

PR #269, opened 2018-10-31
Adapt latest UIKit API (as of Xcode 10.1)

rikner:
Fixes # 

**Type of change:** 

### DIFF
--- a/Sources/CALayer+ContentsGravity.swift
+++ b/Sources/CALayer+ContentsGravity.swift
@@ -6,12 +6,19 @@
 //  Copyright © 2018 flowkey. All rights reserved.
 //
 
-
-extension CALayer {
-    internal enum ContentsGravity: String {
-        case left, center, right, top, bottom, topLeft, topRight
-        case resize, resizeAspectFill, resizeAspectFit = "resizeAspect"
-    }
+public struct CALayerContentsGravity: Equatable {
+    public static let bottom = CALayerContentsGravity()
+    public static let bottomLeft = CALayerContentsGravity()
+    public static let bottomRight = CALayerContentsGravity()
+    public static let center = CALayerContentsGravity()
+    public static let left = CALayerContentsGravity()
+    public static let resize = CALayerContentsGravity()
+    public static let resizeAspect = CALayerContentsGravity()
+    public static let resizeAspectFill = CALayerContentsGravity()
+    public static let right = CALayerContentsGravity()
+    public static let top = CALayerContentsGravity()
+    public static let topLeft = CALayerContentsGravity()
+    public static let topRight = CALayerContentsGravity()
 }
 
 struct ContentsGravityTransformation {
@@ -42,7 +49,7 @@ struct ContentsGravityTransformation {
             return (bounds.height - scaledContents.height) * (1 - layer.anchorPoint.y)
         }
 
-        switch layer.contentsGravityEnum {
+        switch layer.contentsGravity {
         case .resize:
             offset = .zero
             scale = CGSize(width: bounds.width / scaledContents.width, height: bounds.height / scaledContents.height)
@@ -50,7 +57,7 @@ struct ContentsGravityTransformation {
             offset = .zero
             let maxScale = max(bounds.width / scaledContents.width, bounds.height / scaledContents.height)
             scale = CGSize(width: maxScale, height: maxScale)
-        case .resizeAspectFit:
+        case .resizeAspect:
             offset = .zero
             let minScale = min(bounds.width / scaledContents.width, bounds.height / scaledContents.height)
             scale = CGSize(width: minScale, height: minScale)
@@ -74,6 +81,9 @@ struct ContentsGravityTransformation {
             scale = .defaultScale
         case .topRight:
             offset = CGPoint(x: distanceToMaxX, y: distanceToMinY)
+            scale = .defaultScale
+        default:
+            offset = .zero
             scale = .defaultScale
         }
     }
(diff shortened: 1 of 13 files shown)


Sources/CALayer+ContentsGravity.swift
```diff
@@ -6,12 +6,19 @@
 //  Copyright © 2018 flowkey. All rights reserved.
 //
 
-
-extension CALayer {
-    internal enum ContentsGravity: String {
-        case left, center, right, top, bottom, topLeft, topRight
-        case resize, resizeAspectFill, resizeAspectFit = "resizeAspect"
-    }
+public struct CALayerContentsGravity: Equatable {
```
rikner: use `RawRepresentable` instead of Equatable

Sources/CALayer+ContentsGravity.swift
```diff
@@ -75,6 +72,9 @@ struct ContentsGravityTransformation {
         case .topRight:
             offset = CGPoint(x: distanceToMaxX, y: distanceToMinY)
             scale = .defaultScale
+        default:
```
ephemer: we should have no default case here. this will confuse people. I noticed `.bottomLeft` and `.bottomRight` are new: would be great if you could either comment them out for now or implement them here.

Sources/CALayer.swift
```diff
@@ -19,11 +19,7 @@ open class CALayer {
     /// the view sets this value to match the screen.
     open var contentsScale: CGFloat = 1.0
 
-    internal var contentsGravityEnum: ContentsGravity = .resize
-    open var contentsGravity: String {
-        get { return contentsGravityEnum.rawValue }
-        set { contentsGravityEnum = ContentsGravity(rawValue: newValue) ?? .center } // matches iOS
-    }
+    open var contentsGravity: CALayerContentsGravity = .resize
```
ephemer: I think this `.center` fallback behaviour is still relevant, probably in `CALayerContentsGravity(rawValue: String)`

Sources/CALayer+ContentsGravity.swift
```diff
@@ -1,17 +1,14 @@
...
-        case left, center, right, top, bottom, topLeft, topRight
-        case resize, resizeAspectFill, resizeAspectFit = "resizeAspect"
-    }
+/*
+ * Apples CALayerContentsGravity implementation is based on a struct
+ * with a raw representable because of backwards compatibility
+ * We implemented it with an enum, which can be used the
+ * same way as Apples CALayerContentsGravity
+*/
+public enum CALayerContentsGravity {
```
ephemer: This should be `public enum CALayerContentsGravity: String {` to allow initing via `CALayerContentsGravity(rawValue:)` which is still possible in the new Xcode version (via its `RawRepresentable` conformance) -> see note about the `.center` fallback

Sources/CALayer.swift
rikner: like this? feels very strange:
f0b41fe33c9298fda07238706b6022cb351e324a

Sources/CALayer+ContentsGravity.swift
rikner: oh yeah, kind of stupid of me, just wanted a compiling version when I did this ^^

Sources/CALayer.swift
rikner: Since we implement `CALayerContentsGravity` with an `enum` and it's `init(rawValue)` is failable, we don't need the fallback anymore. The API consumer is not able to do anything stupid when setting `CALayer.contentsgravity` via `init(rawValue:)?`, because he has to provide a fallback himself.

Sources/CALayer+ContentsGravity.swift
```diff
@@ -42,15 +39,15 @@ struct ContentsGravityTransformation {
             return (bounds.height - scaledContents.height) * (1 - layer.anchorPoint.y)
         }
 
-        switch layer.contentsGravityEnum {
+        switch layer.contentsGravity {
```
michaelknoch: The name of this was `contentsGravityEnum`? I thought this is an internal type for us

Sources/CALayer.swift
ephemer: You're right, but there may well be users who rely on `init(rawValue:)`, meaning we should implement it too, including the weird fallback. It would be nicer if Apple just used an enum like we did, but it's very likely because their UIKit is in ObjC.

We can (and should) still use an enum ourselves, but with `: String` to add the initializer

Sources/CALayer+ContentsGravity.swift
ephemer: it was, but now UIKit is using an API compatible with ours, so we can remove our `...Enum` implementation

Sources/CALayer+ContentsGravity.swift
```diff
@@ -1,17 +1,14 @@
...
-        case left, center, right, top, bottom, topLeft, topRight
-        case resize, resizeAspectFill, resizeAspectFit = "resizeAspect"
-    }
+/*
+ * Apples CALayerContentsGravity implementation is based on a struct
+ * with a raw representable because of backwards compatibility
+ * We implemented it with an enum, which can be used the
+ * same way as Apples CALayerContentsGravity
+*/
+public enum CALayerContentsGravity: String {
```
ephemer: In an ideal world we'd implement the fallback to `.center` for an unknown value here, like I'm pretty sure is still the case with Apple UIKit. But since it's not documented I don't think it's reasonable for anyone to rely on this behaviour. So I'm fine for this to go in as is!

michaelknoch:
merge? @rikner

PR #270, opened 2018-11-05
(Mostly) Android: Fix crashes and render bugs after recreating GLContext

### LINKED ISSUE #208: SegFault on CGImage deallocation leads to app crashes
**Priority:** High

## Current behavior
When opening the player it crashes or freezes sometimes.

Some Devices where this occurred so far:
- Galaxy Tab A with S Pen - SM-P585M (Android 7.0)
- ONEPLUS A5000 (Android 8.1.0)
- Nexus 7 (Android 5.1.1)
- Huawei Mate 10 Pro (Android 8.0.0)

## Log
[Link to Bugsnag logs](https://app.bugsnag.com/flowkey-gmbh/flowkey/errors/5afd98a191b57f001cccef43?filters[event.since][0]=30d&filters[error.status][0]=open)

```
Segmentation fault: Fatal signal from native code: 11 (Segmentation fault)
        at FreeImage(libSDL2.so:1163688)
        at UIKit.CGImage.__deallocating_deinit(UIKit/Sources/CGImage.swift:51)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4410220)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4399768)
        at UIKit.CALayer.deinit(libUIKit.so:343836)
        at UIKit.CALayer.__deallocating_deinit(UIKit/Sources/CALayer.swift:11)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4410220)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4399768)
        at UIKit.UILabel.deinit(libUIKit.so:662968)
        at UIKit.UILabel.__deallocating_deinit(libUIKit.so:668572)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4409896)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4409464)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
        at swift::Demangle::demangleOldSymbolAsNode(llvm::StringRef, swift::Demangle::NodeFactory&)(libswiftCore.so:5207956)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4399768)
        at FlowkeyPlayer.SongMetaDataView.__deallocating_deinit(libAndroidPlayer.so:595208)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4409896)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4409464)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
        at swift::Demangle::demangleOldSymbolAsNode(llvm::StringRef, swift::Demangle::NodeFactory&)(libswiftCore.so:5207956)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4399768)
        at FlowkeyPlayer.LoadingView.__deallocating_deinit(libAndroidPlayer.so:484068)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4409896)
        at function signature specialization <preserving fragile attribute, Arg[2] = Owned To Guaranteed> of function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._insertionSort<A where A: Swift.BidirectionalCollection, A: Swift.MutableCollection>(inout A, subRange: Swift.Range<A.Index>, by: (A.Element, A.Element) throws -> Swift.Bool) throws -> ()(libswiftCore.so:4409464)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4399768)
        at PlayerTestAppSDL.PlayerLoadingController.__deallocating_deinit(Sources/PlayerTestAppSDL/PlayerLoadingController.swift:36)
        at function signature specialization <preserving fragile attribute, Arg[1] = Owned To Guaranteed> of Swift._introSort<A where A: Swift.MutableCollection, A: Swift.RandomAccessCollection, A.Element: Swift.Comparable>(inout A, subRange: Swift.Range<A.Index>) -> ()(libswiftCore.so:4405300)
```

## User description

> Sometimes when I try to open a song, the app either crashes, or the loading never ends. In the latter situation, I am forced to reload the song.
Sometimes, when I'm playing a song, the video will pause out of nowhere, and it won't unpause for a moment. The video may also skip a few seconds ahead of me and pause, in that case, it will only unpause when I correctly play all the notes up to the moment where it paused.

[intercom conversation link](https://app.intercom.io/a/apps/f716eb3f738e23f5ba4a93502a6b49616ee717a4/inbox/inbox/715521/conversations/16286071032)

## Environment
Reported most recently and frequently here:
- Device: Galaxy Tab A with S Pen (SM-P585M)
- OS Version: Android 7.0
- App Version: Build 1154598

ephemer:
Fixes #208 (and more)

This is a huge improvement to the way rendering works on Android. Until now we had to reload the entire app when the GLContext was killed (as is the case when minimising the app, entering the task switcher etc).

Instead, we now store the image's source data in `CGImage` and use it to re-upload the image data onto the GPU when entering the app again, in the newly recreated GLContext. This means the app and all of its state stays loaded at minimal cost (our PNG sources add up to about 100kb RAM usage). It now looks and feels just like an iOS app 🔥

There are also some related changes to font rendering where statically-stored `UIFont` instances would not render correctly in a newly-created GLContext. `UIFont` is now a little more separated from `FontRenderer`: instead of holding a strong reference to a specific `FontRenderer`, `FontRenderer`s can be recreated behind the scenes and `UIFont` accesses the relevant `FontRenderer` instance dynamically. Aside from the rendering improvements (text stays consistently visible!), this should also fix some crashes when cleaning up fonts.

Finally, we are updating `sdl-gpu` to avoid crashes when "free"ing a GPU_Image that was created in a now-invalid/destroyed `GPU_Renderer` context. This was causing crashes in over 5% of sessions and will be a huge quality of life improvement for Android users.

### Review note:

I tried to make the commit history clean to allow reviewing the different parts of this PR individually, so I can recommend going through each commit one by one.

### Important:

We are now using our own fork of sdl-gpu, so you will probably have to run `git submodule sync` for git to recognise the changed remote.

### DIFF
--- a/Sources/CALayer.swift
+++ b/Sources/CALayer.swift
@@ -241,9 +241,9 @@ open class CALayer {
     internal var _presentation: CALayer?
     open func presentation() -> CALayer? { return _presentation }
 
-    var disableAnimations = false
+    internal var disableAnimations = false
 
-    var animations = [String: CABasicAnimation]() {
+    internal var animations = [String: CABasicAnimation]() {
         didSet { onDidSetAnimations(wasEmpty: oldValue.isEmpty) }
     }
 
@@ -251,6 +251,13 @@ open class CALayer {
     /// This is both a performance optimization (avoids lots of animations at the start)
     /// as well as a correctness fix (matches iOS behaviour). Maybe there's a better way though?
     internal var hasBeenRenderedInThisPartOfOverallLayerHierarchy = false
+
+    internal var _needsDisplay = true
+    open func needsDisplay() -> Bool { return _needsDisplay }
+    open func setNeedsDisplay() { _needsDisplay = true }
+    open func display() {
+        delegate?.display(self)
+    }
 }
 
 private extension CGPoint {
@@ -299,4 +306,5 @@ extension CALayer: Hashable {
 
 public protocol CALayerDelegate: class {
     func action(forKey event: String) -> CABasicAnimation?
+    func display(_ layer: CALayer)
 }
--- a/Sources/FontRenderer+singleLineSize.swift
+++ b/Sources/FontRenderer+singleLineSize.swift
@@ -116,7 +116,7 @@ extension FontRenderer {
             }
 
             guard Find_Glyph(self.rawPointer, characterCode, CACHED_METRICS) == 0 else {
-                assertionFailure("Couldn't find glyph")
+                print("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') could not be found")
                 return nil
             }
 
@@ -124,7 +124,7 @@ extension FontRenderer {
 
             let spaceCharacterCode = 32
             if characterCode != spaceCharacterCode, glyph.maxx - glyph.minx <= 0 {
-                assertionFailure("Glyph has no width")
+                print("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') has no width")
                 return nil
             }
 
--- a/Sources/FontRenderer+singleLineSize.swift
+++ b/Sources/FontRenderer+singleLineSize.swift
@@ -116,7 +116,7 @@ extension FontRenderer {
             }
 
             guard Find_Glyph(self.rawPointer, characterCode, CACHED_METRICS) == 0 else {
-                print("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') could not be found")
+                assertionFailure("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') could not be found")
                 return nil
             }
 
@@ -124,7 +124,7 @@ extension FontRenderer {
 
             let spaceCharacterCode = 32
             if characterCode != spaceCharacterCode, glyph.maxx - glyph.minx <= 0 {
-                print("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') has no width")
+                assertionFailure("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') has no width")
                 return nil
             }
 
--- a/Sources/UIApplication+handleSDLEvents.swift
+++ b/Sources/UIApplication+handleSDLEvents.swift
@@ -67,7 +67,8 @@ extension UIApplication {
 
                 if keyModifier.contains(KMOD_LGUI) || keyModifier.contains(KMOD_RGUI) {
                     if e.key.keysym.sym == 114 { // CMD-R
-                        UIApplication.restart()
+                        UIScreen.main = nil
+                        UIScreen.main = UIScreen()
                     }
                 }
                 #endif
@@ -93,29 +94,29 @@ extension UIApplication {
 
 extension UIApplication {
     static func onWillEnterForeground() {
-        UIApplication.restart {
-            if let runningApplication = UIApplication.shared {
-                runningApplication.delegate?.applicationWillEnterForeground(runningApplication)
-            }
+        #if os(Android)
+        if UIScreen.main == nil { // sometimes we "enter foreground" after just a loss of focus
+            UIScreen.main = UIScreen()
         }
+        #endif
+
+        UIApplication.shared?.delegate?.applicationWillEnterForeground(UIApplication.shared)
     }
 
     static func onDidEnterForeground() {
-        if let runningApplication = UIApplication.shared {
-            runningApplication.delegate?.applicationDidBecomeActive(runningApplication)
-        }
+        UIApplication.shared?.delegate?.applicationDidBecomeActive(UIApplication.shared)
     }
 
     static func onWillEnterBackground() {
-        if let runningApplication = UIApplication.shared {
-            runningApplication.delegate?.applicationWillResignActive(runningApplication)
-        }
+        UIApplication.shared?.delegate?.applicationWillResignActive(UIApplication.shared)
     }
 
     static func onDidEnterBackground() {
-        if let runningApplication = UIApplication.shared {
-            runningApplication.delegate?.applicationDidEnterBackground(runningApplication)
-        }
+        UIApplication.shared?.delegate?.applicationDidEnterBackground(UIApplication.shared)
+
+        #if os(Android)
+        UIScreen.main = nil
+        #endif
     }
 }
 
--- a/Sources/UIScreen+render.swift
+++ b/Sources/UIScreen+render.swift
@@ -1,0 +1,132 @@
+//
+//  UIScreen+render.swift
+//  UIKit
+//
+//  Created by Chris on 08.08.17.
+//  Copyright © 2017 flowkey. All rights reserved.
+//
+
+import SDL
+import SDL_gpu
+import func Foundation.round
+
+extension UIScreen {
+    func render(window: UIWindow?, atTime frameTimer: Timer) {
+        guard let window = window else {
+            print("Not rendering because `window` was `nil`")
+            return
+        }
+
+        DisplayLink.activeDisplayLinks.forEach { $0.callback() }
+        UIView.animateIfNeeded(at: frameTimer)
+        // XXX: It's possible for drawing to crash if the context is invalid:
+        window.sdlDrawAndLayoutTreeIfNeeded()
+
+        guard CALayer.layerTreeIsDirty else {
+            // Nothing changed, so we can leave the existing image on the screen.
+            return
+        }
+
+        // Layer tree can be made dirty again in layer.sdlRender
+        // So set this here and only reset it if the .flip fails
+        CALayer.layerTreeIsDirty = false
+
+        self.clear()
+        GPU_MatrixMode(GPU_MODELVIEW)
+        GPU_LoadIdentity()
+
+        self.clippingRect = window.bounds
+        window.layer.sdlRender()
+
+        do {
+            try self.flip()
+        } catch {
+            CALayer.layerTreeIsDirty = true
+            assertionFailure("UIScreen failed to render. This shouldn't happen anymore since we added more error handling when rendering the layer tree! Error: \(error)")
+        }
+    }
+
+    func blit(
+        _ image: CGImage,
+        anchorPoint: CGPoint,
+        contentsScale: CGFloat,
+        contentsGravity: ContentsGravityTransformation,
+        opacity: Float
+    ) throws {
+        GPU_SetAnchor(image.rawPointer, Float(anchorPoint.x), Float(anchorPoint.y))
+        GPU_SetRGBA(image.rawPointer, 255, 255, 255, opacity.normalisedToUInt8())
+
+        GPU_BlitTransform(
+            image.rawPointer,
+            nil,
+            self.rawPointer,
+            Float(contentsGravity.offset.x),
+            Float(contentsGravity.offset.y),
+            0, // rotation in degrees
+            Float(contentsGravity.scale.width / contentsScale),
+            Float(contentsGravity.scale.height / contentsScale)
+        )
+
+        try throwOnErrors(ofType: [GPU_ERROR_USER_ERROR])
+    }
+
+    func setShapeBlending(_ newValue: Bool) {
+        GPU_SetShapeBlending(newValue)
+    }
+
+    func setShapeBlendMode(_ newValue: GPU_BlendPresetEnum) {
+        GPU_SetShapeBlendMode(newValue)
+    }
+
+    func clear() {
+        GPU_Clear(rawPointer)
+    }
+
+    func fill(_ rect: CGRect, with color: UIColor, cornerRadius: CGFloat) {
+        if cornerRadius >= 1 {
+            GPU_RectangleRoundFilled(rawPointer, rect.gpuRect(scale: scale), cornerRadius: Float(cornerRadius), color: color.sdlColor)
+        } else {
+            GPU_RectangleFilled(rawPointer, rect.gpuRect(scale: scale), color: color.sdlColor)
+        }
+    }
+
+    func outline(_ rect: CGRect, lineColor: UIColor, lineThickness: CGFloat) {
+        GPU_SetLineThickness(Float(lineThickness))
+        GPU_Rectangle(rawPointer, rect.gpuRect(scale: scale), color: lineColor.sdlColor)
+    }
+
+    func outline(_ rect: CGRect, lineColor: UIColor, lineThickness: CGFloat, cornerRadius: CGFloat) {
+        if cornerRadius > 1 {
+            GPU_SetLineThickness(Float(lineThickness))
+            GPU_RectangleRound(rawPointer, rect.gpuRect(scale: scale), cornerRadius: Float(cornerRadius), color: lineColor.sdlColor)
+        } else {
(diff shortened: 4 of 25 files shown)


Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -67,7 +67,8 @@ extension UIApplication {
 
                 if keyModifier.contains(KMOD_LGUI) || keyModifier.contains(KMOD_RGUI) {
                     if e.key.keysym.sym == 114 { // CMD-R
-                        UIApplication.restart()
+                        UIScreen.main = nil
+                        UIScreen.main = UIScreen()
```
michaelknoch: Does it make a difference if we remove `UIScreen.main = nil`. Setting a new instance instantly should also deinit the old uiscreen instance

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -93,29 +94,29 @@ extension UIApplication {
...
     static func onWillEnterForeground() {
-        UIApplication.restart {
-            if let runningApplication = UIApplication.shared {
-                runningApplication.delegate?.applicationWillEnterForeground(runningApplication)
-            }
+        #if os(Android)
+        if UIScreen.main == nil { // sometimes we "enter foreground" after just a loss of focus
+            UIScreen.main = UIScreen()
         }
+        #endif
```
michaelknoch: why is this android only?

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -93,29 +94,29 @@ extension UIApplication {
...
     }
 
     static func onDidEnterBackground() {
-        if let runningApplication = UIApplication.shared {
-            runningApplication.delegate?.applicationDidEnterBackground(runningApplication)
-        }
+        UIApplication.shared?.delegate?.applicationDidEnterBackground(UIApplication.shared)
+
+        #if os(Android)
+        UIScreen.main = nil
```
michaelknoch: woulnd't it make sense to remove the instance for mac, too?

Sources/UIScreen+render.swift
```diff
@@ -0,0 +1,132 @@
...
+//  Created by Chris on 08.08.17.
+//  Copyright © 2017 flowkey. All rights reserved.
+//
+
+import SDL
+import SDL_gpu
+import func Foundation.round
+
+extension UIScreen {
+    func render(window: UIWindow?, atTime frameTimer: Timer) {
```
michaelknoch: I asked myself if it makes sense to pass window as optional. Maybe we unwrap it in UIApplication and call render only when window exists.

Sources/UIScreen.swift
```diff
@@ -10,16 +10,18 @@ import SDL
...
 import func Foundation.round
 
-
 extension SDLWindowFlags: OptionSet {}
 
 public extension UIScreen {
-    // Default value for testing. Gets set in TestSetup.swift
-    internal(set) public static var main: UIScreen!
+    internal(set) public static var main: UIScreen! {
+        didSet { CALayer.layerTreeIsDirty = true }
```
michaelknoch: I ask myself more often if CAlayer is the right namespace for `layerTreeIsDirty`. Maybe it would be clearer if we move it to UIScreen? ... proably itchy

Sources/UIScreen.swift
```diff
@@ -10,16 +10,18 @@ import SDL
...
+    }
 }
 
 public final class UIScreen {
-    private var rawPointer: UnsafeMutablePointer<GPU_Target>!
+    // If we could use `private` members from an extension this would be private
+    // Keep that in mind when using it: i.e. if possible, don't ;)
+    internal var rawPointer: UnsafeMutablePointer<GPU_Target>!
```
michaelknoch: maybe this is a viable case for prefixing with `_`

Sources/UIApplication+handleSDLEvents.swift
ephemer: The order is important. We have to deinit the old one first

Sources/UIApplication+handleSDLEvents.swift
ephemer: It's not necessary on Mac.

The fact that it's wrapped in a condition anyway means we *could* do this all the time, but I'd rather keep things working how we expect, based on our intentions. We are killing UIScreen when enteringBackground on Android only - see below in this file - so it only makes sense to expect to have to recreate it on Android.

Sources/UIApplication+handleSDLEvents.swift
ephemer: We could I guess but it's not necessary

Sources/CALayer.swift
```diff
@@ -241,16 +241,23 @@ open class CALayer {
     internal var _presentation: CALayer?
     open func presentation() -> CALayer? { return _presentation }
 
-    var disableAnimations = false
+    internal var disableAnimations = false
 
-    var animations = [String: CABasicAnimation]() {
+    internal var animations = [String: CABasicAnimation]() {
```
michaelknoch: isn't internal already the default access level?

Sources/UIScreen+render.swift
ephemer: Yes I'm not 100% sure on this yet either.

The main thing is that `UIApplication.keyWindow` is optional itself, so it made more sense to me to just pass that variable in, rather than checking every time from the outside. The API's intention is that it can *try* to render the UIApplication's keyWindow (or another if you want) but it shouldn't break otherwise

Sources/UIScreen.swift
ephemer: Maybe, but I noticed when prefixing `UIFont.availableFonts` that putting an underscore means these vars appear *first* in code completion, which is kind of the opposite of what we want. So I think for now it'd be best to just look out for this in code reviews

Sources/FontRenderer+singleLineSize.swift
```diff
@@ -116,15 +116,15 @@ extension FontRenderer {
             }
 
             guard Find_Glyph(self.rawPointer, characterCode, CACHED_METRICS) == 0 else {
-                assertionFailure("Couldn't find glyph")
+                print("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') could not be found")
```
michaelknoch: are you sure we don't want to crash in devel when this happens? It may get lost

Sources/UIScreen.swift
ephemer: Yes I wondered the same but then it's weird for random CALayers to access `UIScreen.main` too. The best option might be to do this via proxy: e.g.

```
// in a layer:
CALayer.invalidateLayerTree()

extension CALayer {
  func invalidateLayerTree() { 
     UIScreen.main?.needsRerender = true
  }
}
```

What do you think about that solution? @michaelknoch

Sources/UIScreen+render.swift
michaelknoch: ok, fair enough

Sources/CALayer.swift
ephemer: It is, but given that we're making a library for public consumption I'd like to spell out `internal` vars/`func`s wherever they occur. Part of that decision is to make people think twice about making these `public`: if someone has written `internal` it makes it clear that they didn't just forget to write `public` – that it's by design. Does that make sense @michaelknoch ?

michaelknoch:
overall that looks awesomde dude! Booting my android device right now for some testing.

Sources/FontRenderer+singleLineSize.swift
ephemer: Good point. I'm fine to make these assert again now that we've fixed this issue. Nicely spotted, thanks @michaelknoch

Sources/UIScreen.swift
michaelknoch: I don't think accessing UIScreen.main from here is that random. Its more random for me that CALayer has a static context at all. I always think layerTreeIsDirty is an instance property and the layer and all its childLayers is dirty. Its kind of stupid because it obviously is a static property but I had this confusion several times already ^^

ephemer: It would be more correct to bubble changes up through the layer tree:

```
extension CALayer {
  private var isDirty = true
  func invalidate() {
    isDirty = true
    superlayer?.invalidate()
  }
}
```

But we don't need the granularity and this would have a high performance cost for no benefit over the current solution (because if one thing is dirty we have to rerender the entire tree anyway)

ephemer: It wouldn't be correct to access `UIScreen.main` from random layers, because there's nothing to say that layer has any association with that specific screen at all. And if we end up supporting multiple "screens" we'd have to do a more major refactor than what we'd have to now.

Thinking about this further, this could make the most sense as a property on `UIApplication.shared.keyWindow`, but again I don't think we have a reasonable way of accessing a view's associated window (yet). Also the window does not get destroyed alongside the GLContext as is, meaning we wouldn't gain much from having this property on the window instance anyway (it would have to be treated as if it was static anyway)

So I think I'm just going to leave this for now. But I agree that it's still an unanswered question in our design.

Sources/UIScreen.swift
ephemer: Hi @mouselangelo, that could be an interesting option if you have an idea how we can avoid warnings popping up all through our code where we are using it legitimately? My impression is we'd need to have a lot of boilerplate at those places to suppress the warnings?

mouselangelo: Yup, I agree. Will add unnecessary boilerplate.

Issue #271, opened 2018-11-29
Implement Notifications for applicationWillResignActive etc.


ephemer:
iOS provides various lifecycle notifications when entering / leaving the background. Let's implement these because they'll be useful for the flowkey app but also for general use.

ephemer:
Fixed by #273

PR #273, opened 2018-11-30
Application state notifications

mouselangelo:
Implements #271 

**Type of change:** Feature

iOS provides application state change notifications via `NotificationCenter`, which can be useful.
Also, including the `NotificationCenter` and related types, it becomes possible for developers to add and post their own custom notifications in their apps.

### DIFF
--- a/Sources/UIApplication+handleSDLEvents.swift
+++ b/Sources/UIApplication+handleSDLEvents.swift
@@ -101,22 +101,22 @@ extension UIApplication {
         #endif
 
         UIApplication.shared?.delegate?.applicationWillEnterForeground(UIApplication.shared)
-        UIApplication.post(UIApplication.willEnterForegroundNotification)
+        UIApplication.post(willEnterForegroundNotification)
     }
 
     static func onDidEnterForeground() {
         UIApplication.shared?.delegate?.applicationDidBecomeActive(UIApplication.shared)
-        UIApplication.post(UIApplication.didBecomeActiveNotification)
+        UIApplication.post(didBecomeActiveNotification)
     }
 
     static func onWillEnterBackground() {
         UIApplication.shared?.delegate?.applicationWillResignActive(UIApplication.shared)
-        UIApplication.post(UIApplication.willResignActiveNotification)
+        UIApplication.post(willResignActiveNotification)
     }
 
     static func onDidEnterBackground() {
         UIApplication.shared?.delegate?.applicationDidEnterBackground(UIApplication.shared)
-        UIApplication.post(UIApplication.didEnterBackgroundNotification)
+        UIApplication.post(didEnterBackgroundNotification)
 
         #if os(Android)
         UIScreen.main = nil
--- a/Sources/UIApplication+handleSDLEvents.swift
+++ b/Sources/UIApplication+handleSDLEvents.swift
@@ -93,7 +93,7 @@ extension UIApplication {
 }
 
 extension UIApplication {
-    static func onWillEnterForeground() {
+    class func onWillEnterForeground() {
         #if os(Android)
         if UIScreen.main == nil { // sometimes we "enter foreground" after just a loss of focus
             UIScreen.main = UIScreen()
@@ -104,17 +104,17 @@ extension UIApplication {
         UIApplication.post(willEnterForegroundNotification)
     }
 
-    static func onDidEnterForeground() {
+    class func onDidEnterForeground() {
         UIApplication.shared?.delegate?.applicationDidBecomeActive(UIApplication.shared)
         UIApplication.post(didBecomeActiveNotification)
     }
 
-    static func onWillEnterBackground() {
+    class func onWillEnterBackground() {
         UIApplication.shared?.delegate?.applicationWillResignActive(UIApplication.shared)
         UIApplication.post(willResignActiveNotification)
     }
 
-    static func onDidEnterBackground() {
+    class func onDidEnterBackground() {
         UIApplication.shared?.delegate?.applicationDidEnterBackground(UIApplication.shared)
         UIApplication.post(didEnterBackgroundNotification)
 
@@ -125,7 +125,7 @@ extension UIApplication {
 }
 
 private extension UIApplication {
-    static func post(_ name: NSNotification.Name) {
+    class func post(_ name: NSNotification.Name) {
         NotificationCenter.default.post(name: name, object: UIApplication.shared)
     }
 }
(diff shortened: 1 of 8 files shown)


Sources/UIApplication.swift
```diff
@@ -97,3 +97,33 @@ public func nativeProcessEventsAndRender(env: UnsafeMutablePointer<JNIEnv>, view
     CFRunLoopRunInMode(kCFRunLoopDefaultMode, max(0.001, remainingFrameTime / 2), true)
 }
 #endif
+
+extension UIApplication {
+    public static let didEnterBackgroundNotification
```
ephemer: Ok interesting, didn't realise these were available publicly on iOS. I noticed on iOS they're implemented as `class let` rather than `static let` -> I think the subtle difference could be that `class let`s can be overridden in a subclass (but I'm not certain about that). Which might let you override the notifications that get sent by the system (for whatever crazy reason).

If that's the case, we should change `.post(UIApplication.didBecomeActiveNotification)` to `.post(didBecomeActiveNotification)` to take possible subclassing into account. We may need to make `onDidEnterBackground` etc. `class` (instead of `static`) functions too for this to work.

To be honest though I'm a bit fuzzy on the minor differences – what I'm saying here may all be meaningless. But I do remember running into this issue with `UIView.layerClass` and from memory changing it to `class var` made the intended behaviour (the ability to change layerClass in subclasses) possible.

Sources/UIApplication.swift
```diff
@@ -97,3 +97,33 @@ public func nativeProcessEventsAndRender(env: UnsafeMutablePointer<JNIEnv>, view
     CFRunLoopRunInMode(kCFRunLoopDefaultMode, max(0.001, remainingFrameTime / 2), true)
 }
 #endif
+
+extension UIApplication {
```
ephemer: Would be great if you could put both of these extensions into a new file, something like `UIApplication+Notifications.swift`

UIKitTests/UIApplication+NSNotificationTests.swift
```diff
@@ -0,0 +1,45 @@
...
+//  Created by Chetan Agarwal on 30.11.18.
+//  Copyright © 2018 flowkey. All rights reserved.
+//
+
+import XCTest
+@testable import UIKit
+
+class UIApplication_NSNotificationTests: XCTestCase {
+    func testApplicationPostsWillEnterForegroundNotification() {
+        let notificationExpectation = expectation(forNotification: .UIApplicationWillEnterForeground,
```
ephemer: Cool that you wrote tests for this 👍 

Off the top of my head I'm unsure whether these will somehow call `UIApplication.shared` and/or `UIScreen.main` to be created - it would be best if we could avoid that because there's been no need to init our OpenGL rendering context in tests until now - it's costly, performance-wise, and reduces the number of environments we can run the tests in.

Would be great if you could investigate whether that's happening and if so we can find a way to avoid it (we have the ability to make a mock `UIScreen`, for example).

UIKitTests/UIApplication+NSNotificationTests.swift
```diff
@@ -0,0 +1,45 @@
...
+        UIApplication.onWillEnterBackground()
+        wait(for: [notificationExpectation], timeout: 0.1)
+    }
+
+    func testApplicationPostsDidEnterBackgroundNotification() {
+        let notificationExpectation = expectation(forNotification: .UIApplicationDidEnterBackground,
+                                                  object: nil,
+                                                  handler: nil)
+        UIApplication.onDidEnterBackground()
+        wait(for: [notificationExpectation], timeout: 0.1)
```
ephemer: minor nitpick: I thought there was a `wait(for: Expectation)` API that took a single expectation rather than an array? That may be cleaner here. Not a big deal either way though

mouselangelo: The `XCTWaiter` class doesn't define a method to wait for a single expectation.
It's possible to write an `extension` and support this, however in this particular test case - it's also possible to skip assigning the variable altogether and use the function `waitForExpectations(timeout: ...)` - I think is very simple test cases like these - this could be a good approach.

Sources/UIApplication.swift
mouselangelo: Ah! It was my first instinct too! I prefer that as well :)

UIKitTests/UIApplication+NSNotificationTests.swift
ephemer: Oh yeah, cool. That sounds good, I guess even if we expand the tests with more expectations we probably won't need to wait for each expectation with a different timeout etc.

Sources/UIApplication.swift
mouselangelo: Yes, you're right - `class` basically means that sub-classes can "override" the variable or function, while `static` makes that property/function implicitly `final`.
Updating these :)

UIKitTests/UIApplication+NSNotificationTests.swift
mouselangelo: @ephemer I checked this and it doesn't create an instance of `UIApplication`, and also the tests aways implicitly initialize the `UIScreen.main` with a mock `UIScreen` via `UIKitTests\TestSetup.swift`.

It's also why I post the notification using:
`NotificationCenter.default.post(name: name, object: UIApplication.shared)`
since `object` is an optional type  (`object anyObject: Any?`).

Philosophically, this is a kind of a catch-22 - since there are `UIApplication` lifecycle notifications being sent without an actual instance of the application, but I think, this should be acceptable. What are your thoughts about that?

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -101,25 +101,35 @@ extension UIApplication {
...
+        UIApplication.post(didEnterBackgroundNotification)
 
         #if os(Android)
         UIScreen.main = nil
         #endif
     }
 }
 
+private extension UIApplication {
+    static func post(_ name: NSNotification.Name) {
```
ephemer: does it make sense to make these `class func`s too? (also what is currently `static func onDidEnterBackground() {` etc?)

I think we won't win much from making `willEnterForegroundNotification` into `class var`s otherwise

mouselangelo:
@ephemer - Please check this once again.
The `class` approach is definitely more extensible in case something needs to be "overridden" in the future, even if that doesn't seem likely sometimes.

ephemer:
@mouselangelo I agree. LGTM merging

PR #274, opened 2018-12-01
add tests for demo app

michaelknoch:
**Type of change:** fix
- fix UIApplication
- update readme
- update demoapp


samples/getting-started/DemoApp/main.swift
```diff
@@ -9,4 +9,5 @@
 import UIKit
 import Foundation
 
-UIApplicationMain(0, nil, nil, NSStringFromClass(AppDelegate.self))
+var argv = [UnsafeMutablePointer<Int8>]()
+_ = UIApplicationMain(0, &argv, nil, NSStringFromClass(AppDelegate.self))
```
rikner: @michaelknoch what's the reason we now must pass an empty `argv` instead of `nil`?

michaelknoch:
@ephemer any thoughts about this pr?

samples/getting-started/DemoApp/main.swift
michaelknoch: Its because of this deprecation warning
screenshot 2018-12-26 at 13 36 18

ephemer: boah danke, finally!

ephemer: @rikner aside from the deprecation, the original reason I made the change on iOS is because iOS now crashes on this call if you pass `nil`, so we have to pass something. But we can just pass CommandLine.unsafeArgv now

PR #277, opened 2018-12-13
Only delete CGImage contents if we have no source data

ephemer:

If we try to render in the background we are likely to falsely set various `CGImage.contents` to `nil`. This shouldn't happen.


ephemer:
This has been open for over a year and I'm pretty sure the issue we were trying to fix here has been fixed elsewhere so I'm going to close this

PR #282, opened 2019-03-04
Improve Touch (Part 1 - Timestamps)

### LINKED ISSUE #217: Android: Improve touch handling

rikner:
Fixes #217 

**Type of change:** 

SDL's touch handling makes it impossible to get accurate timestamps on events, which breaks our velocity scrolling and causes other weird behaviour.

Also manually tested on my Android phone 👍 


Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -143,3 +150,127 @@ extension SDL_Event {
...
+            if
+                let firstExistingEvent = UIEvent.activeEvents.first,
+                let _ = firstExistingEvent.allTouches?.first(where: {$0.touchId == event.tfinger.fingerId})
+            {
+                // Found a matching event, adding current touch to it and returning
+                firstExistingEvent.allTouches?.insert(newTouch)
+                return firstExistingEvent
+            } else {
+                // No matching event found, creating a new one
+                return UIEvent(touch: newTouch)
```
ephemer: something about this implementation broke last I looked at it - which is why I started implementing proper multitouch.

this is half implementing multitouch because it pays attention to `touchId` - for this to work in a minimal implementation we need to remove that and always assume `touchId == 0`

rikner: @ephemer do you remember what exactly broke? this current state seems totally fine to me, even when attempting to use multitouch. If I make the assumption of `touchId == 0` (e.g. replacing `event.tfinger.fingerId` with `0`) then the sheet goes crazy when using multiple fingers.

ephemer:
@rikner it would be awesome if you could provide an update whether this is still relevant, still builds, and still works as expected. From looking quickly at the code it seems like we should be able to merge this  👍

rikner:
@ephemer Yeah, I will take care of this very very soon.

rikner:
@ephemer merged master and tested this on my phone. velocity scrolling now reliably works and I also compared the CPU usage with the master branch, which was exactly the same 👍

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -25,7 +33,7 @@ extension UIApplication {
             case SDL_MOUSEMOTION:
                 if
                     let event = UIEvent.activeEvents.first,
-                    let touch = event.allTouches?.first(where: { $0.touchId == Int(0) } )
+                    let touch = event.allTouches?.first(where: { $0.touchId == Int(0) })
```
ephemer: how come we have to `Int(0)` here rather than just `== 0`?

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -143,3 +150,122 @@ extension SDL_Event {
         return TimeInterval(self.common.timestamp) / 1000
     }
 }
+
+extension UIEvent {
+    static func from(_ event: SDL_Event) -> UIEvent? {
+        switch SDL_EventType(event.type) {
+        case SDL_FINGERDOWN:
+            // XXX: minimal implementation for single touch
+            guard UIEvent.activeEvents.isEmpty else {
```
ephemer: what happens if we _do_ multitouch here?

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -143,3 +150,122 @@ extension SDL_Event {
...
+                ),
+                timestamp: event.timestampInSeconds
+            )
+
+            return UIEvent(touch: newTouch)
+
+        case SDL_FINGERMOTION:
+            if
+                let firstExistingEvent = UIEvent.activeEvents.first,
+                let matchingTouch = firstExistingEvent.allTouches?.first(where: {$0.touchId == event.tfinger.fingerId})
```
ephemer: nit: formatting `{ $0.touchId` ... `.fingerId })`

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -143,3 +150,122 @@ extension SDL_Event {
...
+                matchingTouch.updateAbsoluteLocation(.from(event.tfinger))
+                return firstExistingEvent
+            } else {
+                return nil
+            }
+
+        case SDL_FINGERUP:
+            if
+                let firstExistingEvent = UIEvent.activeEvents.first,
+                let matchingTouch = firstExistingEvent.allTouches?.first(where: {$0.touchId == event.tfinger.fingerId})
```
ephemer: here as well

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -143,3 +150,122 @@ extension SDL_Event {
...
+    pressure: JavaFloat,
+    timestamp: JavaLong
+) {
+    guard let eventType = SDL_EventType.eventFrom(androidAction: action)
+    else { return }
+
+    var event = SDL_Event(tfinger:
+        SDL_TouchFingerEvent(
+            type: eventType.rawValue,
+            timestamp: UInt32(timestamp), // ensure this is in ms
```
ephemer: these comments are actually TODOs. can you confirm them and if so remove them (would be good to name this input parameter `timestampInMs` if that's the case, for example)

rikner:
@janek 💯

PR #283, opened 2019-03-04
Improve Touch (Part 2 - Multitouch)

rikner:

Adds support for multitouch

Currently the branch causes huge slowdowns with multitouch, we may need to reorganise some code to make this efficient

## ToDo
- [ ] resolve performance issues when using multi touch


ephemer:
@rikner what is the status of this PR? I'm marked to review it but it's missing a description etc. Can you please update it when you get a chance and mark me again for review if it's ready? I'm happy to do some work on this at some point myself too if needed

rikner:
@ephemer In the initial PR (#265) we refactored the timestamps and introduced multitouch, but had  performance issues. I split the initial PR into #282 and this one. Currently the performance issues still exist when using multiple fingers. That's the most important issue we should solve.

rikner:
@ephemer 
@janek and me merged the latest master and ran this on a device to reproduce the performance issue. As far as I remember the player got unresponsive for a while when using multiple fingers, but that's not the case anymore. We also looked at the Profiler in Android Studio and saw that the CPU usage was not different from when using single vs multi-touch. Could it be that this has been solved by some other PR?

ephemer:
Hey @rikner thanks for the update here! I would like to try it especially on a device with weak performance to see, it should be immediately obvious whether there are still issues here.

AFAIK we haven't changed anything in another PR so I'd be kind of surprised if the issues have spontaneously gone away. Also curious as to whether you tried this with our app or with the test app (which might hint whether the performance issues are affected by a larger / deeper view hierarchy). Feel free to ping me when you have time and we can have a look at this together

rikner:
@ephemer @janek and me also tested it on the rather crappy samsung phone we have in the office. we did not see anything suspicious in the CPU usage. But let's compare the CPU usage between this branch and `master` again. I'll get to you with that.

PR #284, opened 2019-03-11
Always layout scroll indicators.

janek:
We are considering sacrificing a small, sacrificable amount of performance to have safer and simpler code by laying out scroll indicators always (and not only when they are visible).


janek:
This changes the assumptions I made when writing tests, so I will need to update them before it passes. But I wanted to get first get an opinion on whether we're going through with this or taking another path.

PR #285, opened 2019-04-23
Add codeowners

cshg:
**Type of change:** Dev workflows

We want to use codeowners to protect critical code. In order to change those parts of the code
a review from at least one "owner" is required.
We don't want to introduce unnecessary overheaded by defining overall codeowners. At some point in the future (once the team grows) we might add this for additional safety.

Once this PR is merged we will add the following branch protection rule for master:
image

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,11 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# We use codeowners to protect critical code. In order to change those parts of the code
+# a review from at least one "owner" is required.
+
+Sources/CALayer+SDL.swift @ephemer
+Sources/UIView+SDL.swift @ephemer
+Sources/FontRender*.swift @ephemer
+Sources/Shader*.swift @ephemer
+


michaelknoch:
I would suggest myself as a codeowner for the animation code. But unfortunately all the animation files are in `/sources`

Screenshot 2019-04-23 at 10 12 05

cshg:
@michaelknoch yea, that's a recurring issue that XCode folder structure and file system structure are out of sync :/
Are there some core files we would want to protect, e.g. `*+animations.swift` files? And we add the other ones which are really needed line by line?
I wouldn't do sth like `*Animation*`, that would be too general and could apply somewhere where it's not intended.

michaelknoch:
in general i have the feeling that it may be hard in the future to maintain the codeowner file if we define super specific rules. We could also move all animation related files into an animation subdirectory and reference the directory from the codeowners file. This seems to be a bigger effort for now, but maybe it reduces future pain. 

what do you think @cshg ?

cshg:
@michaelknoch I agree, that moving the Animation specific files in a certain directory in the file structure would be the best for this case. If we put in the effort to do that I would suggest to also do the same for the SDL/rendering related files.

ephemer:
I like the idea of splitting it up but if we do that it should be motivated by project structural clarity and not by codeowners :) but hopefully it’ll be both in one

cshg:
@ephemer agree, codeowners is just one additional benefit/motivation to introduce that structural clarity. Since we currently rely on the XCode folder structure the need for that just hasn't been as strong.

PR #291, opened 2019-08-08
[swift5] fix demo app

michaelknoch:
**Type of change:** fix

fixes the demoapp


samples/getting-started/cmake_caches.cmake
```diff
@@ -0,0 +1,63 @@
+# Copied from NDK's android.toolchain.cmake:
```
michaelknoch: i think this file could exist in `swift-android-toolchain` @ephemer

samples/getting-started/go.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
michaelknoch: we could also move a more abstract version of this build script to the toolchain or at least uikit

samples/getting-started/cmake_caches.cmake
ephemer: Yeah it probably should

samples/getting-started/go.sh
ephemer: yeah I don't see any reason not to put this into the toolchain too!

samples/getting-started/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+cmake_minimum_required(VERSION 3.15.0)
```
ephemer: This should be pretty much the only thing that needs to change here :)

PR #299, opened 2020-01-10
Swift5

### LINKED ISSUE #250: Build Swift for x86 Android (also to support Android Emulator)
## Motivation 

At the moment we can only build for `armv7a`. While this covers the great proportion of Swift devices on the market today, it'd be great to build for x86 devices as well for two main reasons:

- x86 Devices make up about 10% of the total Android market, which is about 200 million devices.
- The Android Emulator only really works with x86-built libraries. The ARM CPU emulation is so slow that it's basically unusable. This isn't a huge issue, but would be interesting particularly for things like automated tests.

## Proposed solution

"Somebody" needs to make a PR on Apple's Swift repo adding support for Android x86. I suspect this won't be anywhere near as involved as the original port, but it would be a non-trivial piece of work that as far as I know has not been attempted yet.

## Potential alternatives

- Wait for x86 devices to lose market share (if this happens to be the trend over time?)

michaelknoch:
fixes #250

**Type of change:** tooling

- update toolchain and overall structure to support swift5
- this enables builds for x86 and arm64-8a

more info here https://github.com/flowkey/swift-android-toolchain/pull/6

## Todo
- [x] fix ci
- [x] merge toolchain and reference master
- [x] merge SDL and reference master
- [x] merge jni and reference devel


samples/getting-started/android/app/build.gradle
```diff
@@ -19,8 +16,20 @@ android {
...
+            reset()
+            enable true
+            universalApk false
+            include "armeabi-v7a"
+        }
+    }
+    externalNativeBuild {
+        cmake {
+            version "3.16.2"
+            path "CMakeLists.txt"
```
ephemer: not sure we need this?

michaelknoch: pretty sure, i removed path first and then the entire block. In both cases gradle builds the app without any native stuff

PR #304, opened 2020-02-06
explicitly set base sdk and supported platforms for UIKit mac target

michaelknoch:
**Type of change:** dx

This pr explicitly sets base sdk and supported platforms for the mac target to mac osx.
Otherwise apples new build system gets confused when building a project which has a reference to UIKit-cross-platform. My guess is that this is a bug in the build system, because the values should already default to macos (thats at least what xcode is showing) but for some reason they fallback to ios which obviously makes no sense.

### DIFF
--- a/UIKit.xcodeproj/project.pbxproj
+++ b/UIKit.xcodeproj/project.pbxproj
@@ -1379,6 +1379,7 @@
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = macosx;
+				SUPPORTED_PLATFORMS = macosx;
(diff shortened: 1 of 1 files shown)


UIKit.xcodeproj/project.pbxproj
```diff
@@ -1538,6 +1544,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.christianhaug.iOSTestTarget;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
+				SUPPORTED_PLATFORMS = "iphonesimulator iphoneos";
```
michaelknoch: ios test target

PR #305, opened 2020-02-12
Fix GestureRecognizers not filtering `touchesEnded` events properly

janek:
**Type of change:** Bug fix

When a user drags around a view with an active `UIGestureRecognizer`, that recognizer should cancel the touches sent to this view by default (and let them pass if `cancelsTouchesInView` was explicitly set to `false`). 

This PR improves the handling of UITouch's internal `hasBeenCancelledByAGestureRecognizer` variable to ensure that happens.

### DIFF
--- a/Sources/UITouch.swift
+++ b/Sources/UITouch.swift
@@ -53,6 +53,8 @@ public class UITouch {
             action(recognizer)
         }
     }
+
+    var hasBeenCancelledByAGestureRecognizer = false
 }
 
 extension UITouch: Hashable {
--- a/Sources/UITouch.swift
+++ b/Sources/UITouch.swift
@@ -54,7 +54,7 @@ public class UITouch {
         }
     }
 
-    var hasBeenCancelledByAGestureRecognizer = false
+    internal var hasBeenCancelledByAGestureRecognizer = false
 }
 
 extension UITouch: Hashable {
--- a/UIKitTests/Gestures/TouchHandlingTests.swift
+++ b/UIKitTests/Gestures/TouchHandlingTests.swift
@@ -48,6 +48,7 @@ class TouchHandlingTests: XCTestCase {
         handleTouchUp(CGPoint(x: 15, y: 10))
 
         XCTAssertFalse(view.touchesMovedWasCalled)
+        XCTAssertFalse(view.touchesEndedWasCalled)
     }
 
     func testDoesNotCancelTouchesInView() {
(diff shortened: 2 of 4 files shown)


Sources/UITouch.swift
```diff
@@ -53,6 +53,8 @@ public class UITouch {
             action(recognizer)
         }
     }
+
+    var hasBeenCancelledByAGestureRecognizer = false
```
ephemer: please add an explicit `internal` here, as in `internal var` - we are always (hopefully) explicit about this in the UIKit project to avoid confusion and make our intent clear

janek: done!

UIKitTests/Gestures/TouchHandlingTests.swift
```diff
@@ -80,9 +82,15 @@ class TouchHandlingTests: XCTestCase {
 private extension TouchHandlingTests {
     class ResponderView: UIView {
         var touchesMovedWasCalled = false
+        var touchesEndedWasCalled = false
+
         override func touchesMoved(_ touches: Set<UITouch>, with event: UIEvent?) {
             touchesMovedWasCalled = true
         }
+
+        override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent?) {
```
ephemer: Hey @janek, cool that you've expanded this test.

It would be great if you could add a test for the two boolean values (i.e. when `cancelsTouchesInView` is `true` – the default – but also when it's `false`)

janek: thanks! 

that's actually there, we had preexisting `testCancelsTouchesInView()` and `testDoesNotCancelTouchesInView()` (above). They were both only testing if `touchesMoved` was called (or not). My change is to extend that to test for `touchesEnded` aswell.

I was only asking myself if maybe there is another `touchesSomething` to test for, but I looked now and there's only `Began`. I assume it should be always called, regardless of the `cancelsTouchesInView` value. I'm happy to add it there for completeness if that makes sense!

ephemer: From memory, `touchesBegan` works slightly differently on iOS. If the gesture recogniser fails or doesn't exist then (and only then) will `touchesBegan` be called. So I don't think we can test this in the same way. If you like you can make an issue for that but let's not hold up this bug fix for that reason

ephemer: Ah ok sorry I missed the `XCAssert*False*` in the initial case. Cool then this looks fine to me! Would be great if you could make an issue for the other bit, I'm pretty sure we're not doing it 100% "correctly" right now though

janek: cool, I'm happy to!

PR #307, opened 2020-02-13
Fix scrollview deceleration reporting

janek:
**Type of change:** Bug fix

When comparing behaviour with iOS, I realized that the `isDecelerating` value for `scrollView` is not being reported the same. 

The scenario was: one "fast" pan that starts a velocity scroll, then another touch that stops it, maybe pans around a bit, and releases. 

As soon as the first pan is released and the velocity scroll starts, `isDecelerating` is `true`. 
Then when the second touch starts, the visible deceleration stops and we are reporting `isDecelerating == false`. This seems reasonable, but it's not what iOS does. On iOS, the scrollView is still marked as decelerating, until the touch is released. 

This PR aligns behavior with iOS and adds a test that covers this case.

## Warning
I realized that the difference in behaviour could come from the LongPressRecognizer that's there in iOS (but not yet in UIKit-crossplatform). If that's the case, this is likely not the right fix - please keep that in mind and let me know.

### DIFF
--- a/Sources/UIPanGestureRecognizer.swift
+++ b/Sources/UIPanGestureRecognizer.swift
@@ -11,8 +11,9 @@ import Foundation
 open class UIPanGestureRecognizer: UIGestureRecognizer {
     private var initialTouchPoint: CGPoint?
 
-    private var previousTouchesMovedTimestamp: TimeInterval?
-    private var touchesMovedTimestamp: TimeInterval?
+    // TODO: make an extension in tests instead of changing this to internal from private
+    internal var previousTouchesMovedTimestamp: TimeInterval?
+    internal var touchesMovedTimestamp: TimeInterval?
 
     private let minimumTranslationThreshold: CGFloat = 5
 
--- a/Sources/UIPanGestureRecognizer.swift
+++ b/Sources/UIPanGestureRecognizer.swift
@@ -11,7 +11,6 @@ import Foundation
 open class UIPanGestureRecognizer: UIGestureRecognizer {
     private var initialTouchPoint: CGPoint?
 
-    // TODO: make an extension in tests instead of changing this to internal from private
     internal var previousTouchesMovedTimestamp: TimeInterval?
     internal var touchesMovedTimestamp: TimeInterval?
 
--- a/UIKitTests/UIScrollViewTests.swift
+++ b/UIKitTests/UIScrollViewTests.swift
@@ -110,6 +110,28 @@ class UIScrollViewTests: XCTestCase {
         // XXX: set style, test bgr -> does this need to be tested?
     }
 
+    func testIsDecelerating() {
+        XCTAssertEqual(scrollView.isDecelerating, false)
+
+        let mockTouch = UITouch(touchId: 0, at: CGPoint(x: 0, y: 0), timestamp: 0)
+        scrollView.panGestureRecognizer.trackedTouch = mockTouch
+        scrollView.panGestureRecognizer.touchesBegan([mockTouch], with: UIEvent())
+        mockTouch.updateAbsoluteLocation(CGPoint(x: 100, y: 100))
+        scrollView.panGestureRecognizer.touchesMoved([mockTouch], with: UIEvent())
+
+        // Necessary to mock velocity scroll
+        scrollView.contentOffset = CGPoint(x: -30, y: -40)
+        scrollView.panGestureRecognizer.previousTouchesMovedTimestamp = 3.141
+        scrollView.panGestureRecognizer.touchesMovedTimestamp = 3.140
+
+        XCTAssertEqual(scrollView.isDecelerating, false)
+        scrollView.panGestureRecognizer.touchesEnded([mockTouch], with: UIEvent())
+        XCTAssertEqual(scrollView.isDecelerating, true)
+
+        // XXX: optional and problematic: add another touch that "stops" the velocity scroll
+        // First investigate if it's necessary - maybe this test already covers the problem
+    }
+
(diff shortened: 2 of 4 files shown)


UIKitTests/UIScrollViewTests.swift
```diff
@@ -158,11 +194,11 @@ class UIScrollViewTests: XCTestCase {
...
-    func mockTouch(toPoint: CGPoint, inScrollView scrollView: UIScrollView) {
+    func mockTouch(toPoint point: CGPoint, inScrollView scrollView: UIScrollView) {
         let mockTouch = UITouch(touchId: 0, at: CGPoint(x: 0, y: 0), timestamp: 0)
 
         scrollView.panGestureRecognizer.touchesBegan([mockTouch], with: UIEvent())
-        mockTouch.updateAbsoluteLocation(CGPoint(x: 100, y: 100))
+        mockTouch.updateAbsoluteLocation(point)
         scrollView.panGestureRecognizer.touchesMoved([mockTouch], with: UIEvent())
         scrollView.panGestureRecognizer.touchesEnded([mockTouch], with: UIEvent())
     }
```
janek: This is unrelated to the main purpose of the PR, I just noticed this function ignores the parameter it takes

Sources/UIPanGestureRecognizer.swift
```diff
@@ -11,8 +11,8 @@ import Foundation
 open class UIPanGestureRecognizer: UIGestureRecognizer {
     private var initialTouchPoint: CGPoint?
 
-    private var previousTouchesMovedTimestamp: TimeInterval?
-    private var touchesMovedTimestamp: TimeInterval?
+    internal var previousTouchesMovedTimestamp: TimeInterval?
+    internal var touchesMovedTimestamp: TimeInterval?
```
janek: Is there a way of avoiding changing those to `internal`? The reason I wanted to access them was to simulate a velocity scroll in `testIsDecelerating`

UIKitTests/UIScrollViewTests.swift
```diff
@@ -110,6 +110,42 @@ class UIScrollViewTests: XCTestCase {
         // XXX: set style, test bgr -> does this need to be tested?
     }
 
+    func testIsDecelerating() {
+        // This tests an edge case in iOS behaviour (while also testing some basics along the way)
```
ephemer: "while also testing some basics along the way" is probably not the most maintainable approach. if we have basics to test it'd be best to make basic tests for them :)

ephemer: The test is quite long as is and I find it pretty hard to understand what it's actually testing

janek:
I looked into doing this as a UI test on both iOS and Mac, and it won't be viable for this test, as it won't let us check `isDecelerating` value during a touch (after touch down and before touch up). I also don't see a way of mocking touches for this as a unit test in `iOSTestTarget` - certainly the way it's mocked now doesn't work there. 

Generally, I see your point, I'll do some more manual testing now to see if another way of proving this comes to mind.

Sources/UIPanGestureRecognizer.swift
ephemer: see comment in the test :)

janek:
This is problematic to cover with tests because we can't evaluate conditions during a touch on iOS (after touch down and before touch up). I got here by debugging with prints but I'm having trouble reproducing that now and I'm not 100% sure of this fix. 

I'll close this unmerged for now.

janek:
If we ever reopen this topic, here are the print logs that should illustrate the problem:
```< We start panning the scrollView with high speed >
scrollViewDidScroll; is decelerating: false
scrollViewDidScroll; is decelerating: false
(…)
scrollViewDidScroll; is decelerating: false
scrollViewDidScroll; is decelerating: false
< We let go of the touch >
scrollViewDidEndDragging; is Decelerating? false, willDecelerate? true
< The scrollView is still in a velocity scroll, so it continues scrolling >
scrollViewDidScroll; is decelerating: true
scrollViewDidScroll; is decelerating: true
(…)
scrollViewDidScroll; is decelerating: true
scrollViewDidScroll; is decelerating: true
< We touch again to stop the scroll>
< We lift our touch >
scrollViewDidEndDragging; is Decelerating? true, willDecelerate? false```

Those are from iOS, so ground truth. The bug would be that `isDecelerating` is `false` on Android in the last line - but as I said, I had some trouble recreating that. 

The code snippet to put in DemoApp to be able to recreate the logs: 
```
class ViewController: UIViewController, UIScrollViewDelegate {
    func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
        return
    }

    let label = UILabel()

    override func viewWillAppear(_ animated: Bool) {
        super.viewWillAppear(animated)

        let scrollView = UIScrollView(frame: CGRect(x: 200, y: 100, width: 300, height: 150))
        scrollView.contentSize = CGSize(width: 6000, height: 300)
        scrollView.backgroundColor = .white

        scrollView.delegate = self
        view.backgroundColor = UIColor(red: 0 / 255, green: 206 / 255, blue: 201 / 255, alpha: 1)
        view.addSubview(scrollView)
    }

    func scrollViewDidEndDragging(_ scrollView: UIScrollView, willDecelerate decelerate: Bool) {
        print("scrollViewDidEndDragging; is Decelerating? \(scrollView.isDecelerating), willDecelerate? \(decelerate)")
    }

    func scrollViewDidScroll(_ scrollView: UIScrollView) {
        print("scrollViewDidScroll; is decelerating: \(scrollView.isDecelerating)")
    }
}
```

PR #309, opened 2020-02-19
Fix for `insertSubview` being off by one

### LINKED ISSUE #308: `insertSubview:at:` seems to be off by one
## Problem
`insertSubviews` seems to be off by one, running the same code on iOS and macOS gives a different view hierarchy. Code to reproduce this in our _DemoApp_ is attached at the bottom of the post, shortened pseudode below.

#### Pseudocode:
``` 
add 3 square views 
insert a blue horizontal view at index #1
``` 

#### Results:

|   iOS    |  macOS    |
|:-----|:----:|
|   Screenshot 2020-02-18 at 14 23 16   |   Screenshot 2020-02-18 at 14 41 18   | 
| Correct, blue view seems to be on the position with index 1 | Incorrect, blue view seems to be on position with index 2 | 

Note: this does not affect inserting at index 0 - that one behaves correctly 

## Snippets for reproduction and debugging
[DemoAppControllerForInsertSubviewsTesting.txt](https://github.com/flowkey/UIKit-cross-platform/files/4220332/DemoAppControllerForInsertSubviewsTesting.txt)
[testScrollIndicatorsHierarchyPosition.txt](https://github.com/flowkey/UIKit-cross-platform/files/4220337/testScrollIndicatorsHierarchyPosition.txt)

janek:
Fixes #308 
**Type of change:** Bug fix

## Motivation

We had a bug in `insertSubview` where we would use expected new positions in subview hierarchy before the hierarchy is updated. This fix uses the old hierarchy. 

Visually in DemoApp it looks correct (where it previously was wrong). I also added a test that would fail on master, but passes here and on iOS.

### DIFF
--- a/Sources/UIView.swift
+++ b/Sources/UIView.swift
@@ -185,6 +185,7 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
     open func insertSubview(_ view: UIView, at index: Int) {
         // XXX: This might not cover all cases yet. Managing these two hierarchies is complex...
+        insertSubviewWithoutTouchingLayer(view, at: index)
         let indexOfViewWeJustPushedForwardInArray = index + 1
         if index == 0 {
             layer.insertSublayer(view.layer, at: 0)
@@ -195,8 +196,6 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
             // We didn't replace any view. Just push the new layer to the end of the sublayers array.
             layer.addSublayer(view.layer)
         }
-
-        insertSubviewWithoutTouchingLayer(view, at: index)
     }
 
     private func safeGetSubview(index: Int) -> UIView? {
--- a/Sources/UIView.swift
+++ b/Sources/UIView.swift
@@ -185,17 +185,16 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
     open func insertSubview(_ view: UIView, at index: Int) {
         // XXX: This might not cover all cases yet. Managing these two hierarchies is complex...
-        insertSubviewWithoutTouchingLayer(view, at: index)
-        let indexOfViewWeJustPushedForwardInArray = index + 1
         if index == 0 {
             layer.insertSublayer(view.layer, at: 0)
-        } else if let currentSubview = safeGetSubview(index: indexOfViewWeJustPushedForwardInArray) {
+        } else if let currentSubview = safeGetSubview(index: index) {
             layer.insertSublayer(view.layer, below: currentSubview.layer)
         } else {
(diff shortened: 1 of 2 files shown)


Sources/UIView.swift
```diff
@@ -185,17 +185,16 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
...
-        } else if let currentSubview = safeGetSubview(index: indexOfViewWeJustPushedForwardInArray) {
+        } else if let currentSubview = safeGetSubview(index: index) {
             layer.insertSublayer(view.layer, below: currentSubview.layer)
         } else {
             // The given index was greater than that of any existing subview, meaning:
             // We didn't replace any view. Just push the new layer to the end of the sublayers array.
             layer.addSublayer(view.layer)
         }
-
+
```
ephemer: here is a random bunch of spaces. please adjust your Xcode settings so it removes whitespace at the end of whitespace only lines.

UIKitTests/UIView/UIViewTests+subviews.swift
```diff
@@ -61,6 +61,20 @@ class UIViewSubviewTests: XCTestCase {
         XCTAssertEqual(view.layer.sublayers!, [subview1.layer, subview2.layer])
     }
 
+    func testInsertSubviewAtIndex() {
```
ephemer: @janek does this also run on the iOS test target?

Sources/UIView.swift
```diff
@@ -185,17 +185,16 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
     open func insertSubview(_ view: UIView, at index: Int) {
         // XXX: This might not cover all cases yet. Managing these two hierarchies is complex...
-        let indexOfViewWeJustPushedForwardInArray = index + 1
         if index == 0 {
             layer.insertSublayer(view.layer, at: 0)
-        } else if let currentSubview = safeGetSubview(index: indexOfViewWeJustPushedForwardInArray) {
+        } else if let currentSubview = safeGetSubview(index: index) {
```
ephemer: cool, makes sense :)

janek:
@ephemer yes it does! thanks!

Sources/UIView.swift
ephemer: @janek would be great if you could fix this before merging

UIKitTests/UIView/UIViewTests+subviews.swift
janek: yes :)

Sources/UIView.swift
janek: agh. weird, the setting seems to be in the correct state. Nevertheless, fixed - sorry.

PR #310, opened 2020-03-30
Fix crash on native touch

rikner:
**Type of change:** 
Bug Fix

Fixes a crash that happens when re-entering from a background state while still touching the screen. The app then crashes because `Java_org_libsdl_app_SDLActivity_onNativeTouchUIKit` is called, but the JNIEnv or the object argument is not yet available again. Checking for `mHasFocus` before calling the `onNativeTouch` fixes the issue.

### DIFF
--- a/src/main/java/org/libsdl/app/SDLOnTouchListener.kt
+++ b/src/main/java/org/libsdl/app/SDLOnTouchListener.kt
@@ -56,11 +56,18 @@ interface SDLOnTouchListener: View.OnTouchListener {
         return true
     }
 
-    fun callOnNativeTouchIfHasFocus(touchDevId: Int, pointerFingerId: Int, action: Int, x: Float, y: Float, p: Float, t: Long) {
+    fun callOnNativeTouchIfHasFocus(
+            touchDevId: Int,
+            pointerFingerId: Int,
+            action: Int,
+            x: Float,
+            y: Float,
+            p: Float,
+            t: Long
+    ) {
         if (this.mHasFocus) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/libsdl/app/SDLOnTouchListener.kt
```diff
@@ -41,19 +42,33 @@ interface SDLOnTouchListener: View.OnTouchListener {
...
+    fun callOnNativeTouchIfHasFocus(
+            touchDevId: Int,
+            pointerFingerId: Int,
+            action: Int,
+            x: Float,
+            y: Float,
+            p: Float,
+            t: Long
+    ) {
+        if (this.mHasFocus) {
```
ephemer: can you add a comment why we're doing this?

ephemer:
What happens if an alert (for permissions etc) is showing?

rikner:
> What happens if an alert (for permissions etc) is showing?

@ephemer behaves like expected, we lose focus and we won't receive touch events on the player view until the permission dialogue goes away and we have focus again.

rikner:
> Would be great if you could add a comment here but this looks good to me. Not sure why this fixes an issue with the JNI though?

To me it just seemed that we're calling JNI functions before we the JNIEnv is back on track after we've lost focus. Not sure if this is expected behaviour by the JNI, probably not. But the error message on bugsnag says exactly this about `onNativeTouch`: <Arg[0] = Dead, Arg[1] = Dead>

PR #312, opened 2020-06-29
update exoplayer to 2.11.7

michaelknoch:
**Type of change:** dependency update

- potentially fixes an ANR
- should smoothen play/pause/seek experience

### DIFF
--- a/src/main/java/org/uikit/VideoJNI.kt
+++ b/src/main/java/org/uikit/VideoJNI.kt
@@ -78,21 +78,12 @@ class AVPlayer(parent: SDLActivity, asset: AVURLAsset) {
                 }
             }
 
-            override fun onPlayerError(error: ExoPlaybackException?) {
-                if (error?.type == ExoPlaybackException.TYPE_SOURCE) {
+            override fun onPlayerError(error: ExoPlaybackException) {
+                if (error.type == ExoPlaybackException.TYPE_SOURCE) {
                     nativeOnVideoSourceError()
                     Log.e("SDL", "ExoPlaybackException occurred")
                 }
             }
-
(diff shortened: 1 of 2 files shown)


mouselangelo:
PS: we should also update the react-native expo player :)

src/main/java/org/uikit/VideoJNI.kt
```diff
@@ -78,21 +78,12 @@ class AVPlayer(parent: SDLActivity, asset: AVURLAsset) {
...
             }
-
-            // not used but necessary to implement EventListener interface:
-            override fun onShuffleModeEnabledChanged(shuffleModeEnabled: Boolean) {}
-            override fun onPlaybackParametersChanged(playbackParameters: PlaybackParameters?) {}
-            override fun onTracksChanged(trackGroups: TrackGroupArray?, trackSelections: TrackSelectionArray?) {}
-            override fun onLoadingChanged(isLoading: Boolean) {}
-            override fun onPositionDiscontinuity(reason: Int) {}
-            override fun onRepeatModeChanged(repeatMode: Int) {}
-            override fun onTimelineChanged(timeline: Timeline?, manifest: Any?, reason: Int) {}
```
rikner, line 95: interesting, do you know if these were removed from the interface or is there some default implementation now?

rikner, line 95: interesting, do you know if these were removed from the interface or is there some default implementation now?

michaelknoch: i had to remove a few of them because they did not exist in the interface and therefore you can't override them. but i noticed that it was not mandatory to provide the empty implementations. maybe they already implement empty functions in an abstract class

PR #318, opened 2021-02-10
fix outline rendering when borderWidth > 1

michaelknoch:
**Type of change:** bugfix

We have to manually make sure the drawn rect does not exceed the bounds of the base rect when setting a borderWidth bigger than 1

Look close to the top left edge of the first exercise part border (orange outline) to see the issue:

### broken:
border-bad_round

border-fucked

### fixed:
border-nice_round

nice_border

### DIFF
--- a/Sources/UIScreen+render.swift
+++ b/Sources/UIScreen+render.swift
@@ -100,15 +100,25 @@ extension UIScreen {
             width: rect.size.width - offset,
             height: rect.size.height - offset
         ).gpuRect(scale: scale)
-        
+
         GPU_SetLineThickness(Float(lineThickness))
         GPU_Rectangle(rawPointer, scaledGpuRect, color: lineColor.sdlColor)
     }
 
     func outline(_ rect: CGRect, lineColor: UIColor, lineThickness: CGFloat, cornerRadius: CGFloat) {
         if cornerRadius > 1 {
+            // we want to render the outline 'inside' the rect rather
+            // than exceeding the bounds when lineThickness is bigger than 1
+            let offset = lineThickness / 2
+            let scaledGpuRect = CGRect(
+                x: rect.origin.x + offset,
+                y: rect.origin.y + offset,
+                width: rect.size.width - offset,
+                height: rect.size.height - offset
+            ).gpuRect(scale: scale)
+
             GPU_SetLineThickness(Float(lineThickness))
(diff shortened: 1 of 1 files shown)


Sources/UIScreen+render.swift
```diff
@@ -91,14 +91,34 @@ extension UIScreen {
...
         if cornerRadius > 1 {
+            // we want to render the outline 'inside' the rect rather
+            // than exceeding the bounds when lineThickness is bigger than 1
+            let offset = lineThickness / 2
+            let scaledGpuRect = CGRect(
+                x: rect.origin.x + offset,
+                y: rect.origin.y + offset,
+                width: rect.size.width - offset,
+                height: rect.size.height - offset
+            ).gpuRect(scale: scale)
```
mouselangelo, line 118: nit:
Just wondering - since this logic is repeated twice - if it would be possible to have it in one place and somehow reuse it?

mouselangelo (approved):
@michaelknoch - this makes sense to me. I added a couple of comments but they are not critical IMO so approving this.

Sources/UIScreen+render.swift
```diff
@@ -91,14 +91,34 @@ extension UIScreen {
     }
 
     func outline(_ rect: CGRect, lineColor: UIColor, lineThickness: CGFloat) {
+        // we want to render the outline 'inside' the rect rather
+        // than exceeding the bounds when lineThickness is bigger than 1
```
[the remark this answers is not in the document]
michaelknoch: IMG_3302 2

When using a bigger lineThickness the rect (grey) which renders the borderWidth gets centered along the original bounds of the base rect (blue). Thats why it makes sense to move by offset `let offset = lineThickness / 2`. But I do not fully get how this works with borderdWidths of 2 without retina. It always looked correct in my experiments.

@mouselangelo

Sources/UIScreen+render.swift
michaelknoch: i tried, but the readability really sucked after putting this into a helper function.

Sources/UIScreen+render.swift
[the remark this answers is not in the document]
mouselangelo: Ah ok - so basically by default the border is "centered" but we want it to offset so that it's "inside". Makes sense now.
Thanks for the explanation - the drawing helps too :)

Sources/UIScreen+render.swift
mouselangelo: No problem :) thanks for trying

PR #319, opened 2021-02-12
only cancel bounds animation when setting contentOffset without animations

michaelknoch:
**Type of change:** bugfix (align with ios behaviour)

Due some recent changes in the coursesv2 player logic it can happen that we sometimes have two sheet animations animating back to the beginning of the song or to the beginning of the loop. It might be possible to clean this up, but from UIKit perspective this should work, because initiating another contentOffset animation should just override the existing one and then either create a new animation starting from `layer.bounds` or `presentation.bounds` depending on `beginFromCurrentState`. This works on iOS and it works in our UIKit when velocity scroll is disabled.

I noticed that we are removing any bounds animations when the contentOffset of a scrollView gets updated. We need this, otherwise there is a weird jump in the sheet when deceleration finishes and exoplayer updates the x position. But we also set the contentOffset when animating back in a loop or at the end of the sheet, in this case it does not make much sense to remove the animations.

I think the fix here is to remove the animations only when setting the contentOffset without animation (when set from exoplayer in our case). 

https://user-images.githubusercontent.com/5617793/107763467-f74eec80-6d2e-11eb-808d-0dba7f2cc61f.mp4

### DIFF
--- a/Sources/CATransaction.swift
+++ b/Sources/CATransaction.swift
@@ -11,7 +11,7 @@ public struct CATransaction {
     private var disableActions_ = false
     private var animationDuration_: CGFloat = CALayer.defaultAnimationDuration
 
-    private static var transactionStack = [CATransaction]()
+    internal static var transactionStack = [CATransaction]()
 
     public static func begin() {
         transactionStack.append(CATransaction())
--- a/Sources/CATransaction.swift
+++ b/Sources/CATransaction.swift
@@ -11,7 +11,7 @@ public struct CATransaction {
     private var disableActions_ = false
     private var animationDuration_: CGFloat = CALayer.defaultAnimationDuration
 
-    internal static var transactionStack = [CATransaction]()
+    internal private(set) static var transactionStack = [CATransaction]()
 
     public static func begin() {
         transactionStack.append(CATransaction())
--- a/Sources/UIScrollView+velocity.swift
+++ b/Sources/UIScrollView+velocity.swift
@@ -62,8 +62,10 @@ extension UIScrollView {
     }
 
     func cancelDecelerationAnimations() {
-        layer.removeAnimation(forKey: "bounds")
-        horizontalScrollIndicator.layer.removeAnimation(forKey: "position")
-        verticalScrollIndicator.layer.removeAnimation(forKey: "position")
+        if !layer.animations.isEmpty {
+            layer.removeAnimation(forKey: "bounds")
+            horizontalScrollIndicator.layer.removeAnimation(forKey: "position")
+            verticalScrollIndicator.layer.removeAnimation(forKey: "position")
+        }
     }
 }
--- a/Sources/UIScrollView.swift
+++ b/Sources/UIScrollView.swift
@@ -23,7 +23,13 @@ open class UIScrollView: UIView {
         get { return bounds.origin }
         set {
             guard newValue != contentOffset else { return }
-            cancelDecelerationAnimations()
+
+            // Cancel deceleration animations only when contentOffset gets set without animations.
+            // Otherwise we might cancel any "bounds" animations which are not iniated from velocity scrolling.
+            if isDecelerating && CATransaction.transactionStack.isEmpty {
+                cancelDecelerationAnimations()
+            }
+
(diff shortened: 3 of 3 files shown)


Sources/UIScrollView+velocity.swift
```diff
@@ -62,8 +62,10 @@ extension UIScrollView {
     }
 
     func cancelDecelerationAnimations() {
-        layer.removeAnimation(forKey: "bounds")
-        horizontalScrollIndicator.layer.removeAnimation(forKey: "position")
-        verticalScrollIndicator.layer.removeAnimation(forKey: "position")
+        if !layer.animations.isEmpty {
```
michaelknoch, line 65: this removes a bunch of noise when debugging the animation code. This function gets called everytime exoplayer updates the sheet (every pixel). Removing animations, even when no animations exist, calls `layer.animations { didSet }` which does some computation. I think we can avoid this here. This might have a minor performance impact, but I found it better DX when debugging the animation code

Sources/CATransaction.swift
```diff
@@ -11,7 +11,7 @@ public struct CATransaction {
     private var disableActions_ = false
     private var animationDuration_: CGFloat = CALayer.defaultAnimationDuration
 
-    private static var transactionStack = [CATransaction]()
+    internal static var transactionStack = [CATransaction]()
```
rikner, line 14: It's totally ok to read `transactionStack` from `internal`, but I guess setting the value should only be allowed for this class. So I think `internal private(set) static var ...` would be more correct.

Sources/UIScrollView+velocity.swift
rikner: it should have a positive impact on performance, no?

michaelknoch: yep, but it should be very minor because nothing is happening really, just a few functions keep calling each other every pixel and then early return at the end because no animations exist at all. @rikner

Sources/CATransaction.swift
michaelknoch: good point, changed it

Sources/UIScrollView+velocity.swift
rikner: ok was just confused because you said "a minor performance impact, *but* ..." 😅

michaelknoch: got you 😄 , i meant it has some positive impact **but** i was mainly doing it for DX

rikner (approved):
lgtm, I even tested this change on the course messaging branch and it fixed the issue for me as well🥇 👍

Sources/UIScrollView.swift
```diff
@@ -23,7 +23,13 @@ open class UIScrollView: UIView {
         get { return bounds.origin }
         set {
             guard newValue != contentOffset else { return }
-            cancelDecelerationAnimations()
+
+            // Cancel deceleration animations only when contentOffset gets set without animations.
+            // Otherwise we might cancel any "bounds" animations which are not iniated from velocity scrolling.
```
ephemer, line 28: a couple of things:
- I'm not a fan of making `transactionStack` more accessible than `private` as it was (I don't want other parts of the code relying on it - especially not something as high level as `UIScrollView`; it will lead to spaghetti code and weird dependencies that are hard to debug)
- After looking at this code and the comments for quite a while I still can't figure out what the "transaction stack is empty" thing is doing in the first place. Can we look to see whether the scrollview's layer has a `bounds` animation on it instead?

ephemer (changes requested):
I can't really request changes since it's already been merged but my gut feeling is this isn't the best solution to the issue (but am also not sure if I've properly understood the issue yet)

michaelknoch:
hoops, sorry for the quick merge @ephemer . I can try your approach and revert the change in another pr

michaelknoch:
I am happy to revert this if there is a better solution (there probably is). But unfortunately we cannot simply check if there is a bounds animation in `layer.animation` because we want to know if the `setContentOffset` call happened within an UIView.animate block or not. It may happen that there are still pending animations on the layer at that time @ephemer

PR #320, opened 2021-02-12
cleanup previous fix for velocity scroll animations

michaelknoch:
**Type of change:** cleanup fix

This change is based on the previous feedback of @ephemer  in https://github.com/flowkey/UIKit-cross-platform/pull/319.
Instead of exposing the static transactionStack we can also just check if there is an animation prototype defined in `UIView.currentAnimationPrototype`. If this value is defined, we are currently in a stack trace of an animation closure.

The fix for the velocity scroll is to cancel all animations only when setting the contentOffset not from within an animation closure.


ephemer (approved):
thanks @michaelknoch !

PR #321, opened 2021-02-18
Explicitly unwrap UIScreen.main before accesing its scale property

rikner:
**Type of change:** 

Since `UIScreen.main` is an implicitly unwrapped optional, it would be safer to check it for nil before using its `scale` property.

### DIFF
--- a/Sources/UIView.swift
+++ b/Sources/UIView.swift
@@ -140,7 +140,10 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
     public init(frame: CGRect) {
         self.layer = type(of: self).layerClass.init()
-        self.layer.contentsScale = UIScreen.main.scale
+        if let uiScreenMain = UIScreen.main {
+            self.layer.contentsScale = uiScreenMain.scale
+        }
+        
         super.init()
 
         self.layer.delegate = self


Sources/UIView.swift
```diff
@@ -140,7 +140,10 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
     public init(frame: CGRect) {
         self.layer = type(of: self).layerClass.init()
-        self.layer.contentsScale = UIScreen.main.scale
+        if let uiScreenMain = UIScreen.main {
```
rikner, line 143: `UIScreen.main` is an implicitly unwrapped optional: https://github.com/flowkey/UIKit-cross-platform/blob/196ab113ab457a2c393689317b9fe6c067c75fd7/Sources/UIScreen.swift#L16

Sources/UIView.swift
```diff
@@ -140,7 +140,10 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
...
+            self.layer.contentsScale = uiScreenMain.scale
+        }
```
michaelknoch, line 145: I agree that this will probably prevent the crash from happening, but this might also break the layouting when `self.layer.contentsScale` has an invalid value when an UIView gets rendered. I think a more sustainable fix would be to ensure UIScreen.main exists before the first view gets rendered.

Did you manage to reproduce the crash locally @rikner ?

rikner: I agree it would break layouting, so this is rather just a hotfix. I couldn't reproduce it locally, but I dont see anything else that could chrash during `UIView.init()`.

michaelknoch: Is there any stacktrace on bugsnag which makes us think that UIView.init() crashes the app? @rikner

rikner: @michaelknoch [this one](https://app.bugsnag.com/flowkey-gmbh/mobile-app/errors/5e83408b045d4a0017329c48?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[os.name][0][type]=eq&filters[os.name][0][value]=android&filters[error][0]=5e83408b045d4a0017329c48):

ephemer: Is it possible we're initing UIScreen, changing the orientation (which may deinit UIScreen temporarily) and then initing it again after? If so can we make sure the deinit-reinit thing happens synchronously? If so we can't crash at that point

ephemer (changes requested):
I am not really happy with this as a solution - this knowledge may help us track down the issue but the suggestion here is not the fix we need: it only delays the issue slightly by not crashing here (but we may well crash milliseconds later on some other access of UIScreen)

We need to figure out why UIScreen is nil and fix that

Issue #314, opened 2020-11-04
if view.superview != nil { view.removeFromSuperview() }


chenyunguiMilook:
https://github.com/flowkey/UIKit-cross-platform/blob/0e44ce485bf0e85fc161e1427970760dc2ff951e/Sources/UIView.swift#L211
should be if view.superview != nil { view.removeFromSuperview() } ?

ephemer:
Thanks that looks correct. Could you make a pull request for this and add a test?

chenyunguiMilook:
No problem!

ephemer:
Thanks, @chenyunguiMilook!

michaelknoch:
closing this because https://github.com/flowkey/UIKit-cross-platform/commit/196ab113ab457a2c393689317b9fe6c067c75fd7 was merged. Thanks @chenyunguiMilook

PR #323, opened 2021-02-26
clamp subview index _after_ removing from superview

rikner:
**Type of change:** 

#314 revealed another bug, where the subview index was not correctly bound-checked and could therefore cause a crash.


ephemer (approved):
Thanks @rikner!

PR #324, opened 2021-03-01
Fix insertSubviewWithoutTouchingLayer()

rikner:
**Type of change:** Bug fix

Addresses a bug, where a views layer was mistakenly removed from the superlayer.

### DIFF
--- a/Sources/UIView.swift
+++ b/Sources/UIView.swift
@@ -208,7 +208,7 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
     private func insertSubviewWithoutTouchingLayer(_ view: UIView, at index: Int) {
         // remove from superview without removing from superlayer
         if let oldSuperview = view.superview {
-            oldSuperview.subviews = oldSuperview.subviews.filter { $0 != view }
+            oldSuperview.subviews.removeAll(where: { $0 == view })
             view.superview = nil
             oldSuperview.setNeedsLayout()
         }
@@ -223,7 +223,7 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
         guard let superview = superview else { return }
         self.layer.removeFromSuperlayer()
 
-        superview.subviews = superview.subviews.filter { $0 != self }
+        superview.subviews.removeAll(where: { $0 == self })
         self.superview = nil
         superview.setNeedsLayout()
     }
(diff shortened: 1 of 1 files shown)


Sources/UIView.swift
```diff
@@ -206,7 +206,13 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
     /// Adds a subview without touching the view's layer or any of its sublayers.
     /// We need to be able to add layers at an index that isn't directly related to its subview index.
     private func insertSubviewWithoutTouchingLayer(_ view: UIView, at index: Int) {
-        if view.superview != nil { view.removeFromSuperview() }
```
rikner, line 209: `removeFromSuperview` called `layer.removeFromSuperlayer()`, which we do not want here.

Sources/UIView.swift
```diff
@@ -206,7 +206,13 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
...
+        // remove from superview without removing from superlayer
+        if let oldSuperview = view.superview {
+            oldSuperview.subviews = oldSuperview.subviews.filter { $0 != view }
+            view.superview = nil
+            oldSuperview.setNeedsLayout()
+        }
```
rikner, line 214: pretty much what happens in `removeFromSuperview`, but without `layer.removeFromSuperlayer()`

michaelknoch: I don't fully get why we should not touch the player in this case. Can you explain that @rikner ?

Sources/UIView.swift
```diff
@@ -206,7 +206,13 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
     /// Adds a subview without touching the view's layer or any of its sublayers.
     /// We need to be able to add layers at an index that isn't directly related to its subview index.
     private func insertSubviewWithoutTouchingLayer(_ view: UIView, at index: Int) {
-        if view.superview != nil { view.removeFromSuperview() }
+        // remove from superview without removing from superlayer
+        if let oldSuperview = view.superview {
+            oldSuperview.subviews = oldSuperview.subviews.filter { $0 != view }
```
michaelknoch, line 211: should we use `oldSuperview.subviews.removeAll(where: { $0 == view })` maybe?

rikner: Yeah, seems to make more sense to me this way. Just copied this line from `removeFromSuperview()` below, so we probably want to adjust this there too.

Sources/UIView.swift
rikner: Do mean the player or the layer? 

So we don't want to touch the **layer** here since this function is called this way ^^ (`insertSubviewWithoutTouchingLayer`). I probably wouldn't have found the bug if not for the function name.

In case you mean **player**... not sure if we could fix it there, but I think fixing it here makes more sense in order to get more close to the iOS behavior.

ephemer (approved):
nice find @rikner !

PR #327, opened 2021-03-17
improve multilineSize for text calculations

michaelknoch:
**Type of change:** layouting fix

I noticed that there is too much space above and below UILabels especially when using bigger font sizes. Removing the hack `numberOfRandomExtraLinesRenderedBySDLTTF` seems to fix it. Maybe the rendering behaviour got fixed by updating `SDLTTF`?

**before:**

Screenshot 2021-03-17 at 19 48 01

Screenshot 2021-03-17 at 19 55 24

**after:**
Screenshot 2021-03-17 at 19 47 15

Screenshot 2021-03-17 at 19 43 06

### DIFF
--- a/Sources/FontRenderer.swift
+++ b/Sources/FontRenderer.swift
@@ -161,11 +161,7 @@ extension FontRenderer {
             tok = firstCharOfNextLine
         } while (tok < end)
 
-        // The sizing using linesCount is actually correct, but it gives a different result compared to
-        // what actually gets rendered. So for now, use this updated figure to have correct sizeThatFits etc:
-        let numberOfRandomExtraLinesRenderedBySDLTTF = 1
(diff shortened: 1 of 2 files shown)


rikner (approved):
Looks like we still have to adjust the test which is failing right now.
But apart from this looks good to me.

Sources/FontRenderer.swift
```diff
@@ -161,11 +161,7 @@ extension FontRenderer {
             tok = firstCharOfNextLine
         } while (tok < end)
 
-        // The sizing using linesCount is actually correct, but it gives a different result compared to
-        // what actually gets rendered. So for now, use this updated figure to have correct sizeThatFits etc:
```
ephemer, line 165: "compared to what actually gets rendered" – sounds like there was a difference between the sizing function and the render function. i imagine that was fixed with an update to SDL_ttf yep

ephemer (approved):

PR #328, opened 2021-03-22
prevent recursion when layouting parent from subview

michaelknoch:
**Type of change:** performance improvement

I initially introduced this "weird bottom top" layout dependency in this commit https://github.com/flowkey/UIKit-cross-platform/commit/3ed2303e53d514a5b36127483bf3bb73c160cf7b. But I realised that this dependency causes expensive recursive relayouts at least in our player at the moment. We added a check to only relayout if the size has changed, but in our player implementation the size will always change (look at the example below)

It happens that `layoutSubviews` of the LearnstepNotification gets called a few hundred times initially and during learning again. The recursion is able to stop at some point, but I don't really know why it stops.

```swift
layoutSubviews {
    button.frame = container // changes bounds
    button.sizeToFit() // changes bounds again

   ....
 
   bodyLabel.bounds.size = CGSize(width: container.width, height: 0)
   bodyLabel.sizeToFit()

}
```

### DIFF
--- a/Sources/CALayer.swift
+++ b/Sources/CALayer.swift
@@ -130,12 +130,12 @@ open class CALayer {
 
     open var bounds: CGRect = .zero {
         willSet(newBounds) {
+            guard newBounds != bounds else { return }
+            onWillSet(keyPath: .bounds)
+
             if previousBounds == nil {
                 previousBounds = bounds
             }
-
-            guard newBounds != bounds else { return }
-            onWillSet(keyPath: .bounds)
         }
     }
 
--- a/Sources/CALayer.swift
+++ b/Sources/CALayer.swift
@@ -130,16 +130,16 @@ open class CALayer {
 
     open var bounds: CGRect = .zero {
         willSet(newBounds) {
-            guard newBounds != bounds else { return }
-            onWillSet(keyPath: .bounds)
-
-            if bounds.size != newBounds.size {
-                // It seems weird to access the superview here but it matches the iOS behaviour
-                (self.superlayer?.delegate as? UIView)?.setNeedsLayout()
+            if previousBounds == nil {
+                previousBounds = bounds
             }
 
+            guard newBounds != bounds else { return }
+            onWillSet(keyPath: .bounds)
         }
     }
+
+    internal var previousBounds: CGRect?
 
     public var opacity: Float = 1 {
         willSet(newOpacity) {
--- a/Sources/UIView.swift
+++ b/Sources/UIView.swift
@@ -245,7 +245,6 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
         parentViewController?.viewWillLayoutSubviews()
         parentViewController?.viewDidLayoutSubviews()
-
     }
 
 
--- a/Sources/UIView.swift
+++ b/Sources/UIView.swift
@@ -239,7 +239,7 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
         needsLayout = false
         
         if layer.previousBounds != nil && layer.previousBounds != layer.bounds {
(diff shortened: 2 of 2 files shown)


michaelknoch:
@rikner @ephemer Do you remember why we added this? I remember this being an issue with some layouts with the MIDI settings view maybe? Not entirely sure.

Sources/CALayer.swift
```diff
@@ -132,12 +132,6 @@ open class CALayer {
         willSet(newBounds) {
             guard newBounds != bounds else { return }
             onWillSet(keyPath: .bounds)
-
-            if bounds.size != newBounds.size {
-                // It seems weird to access the superview here but it matches the iOS behaviour
-                (self.superlayer?.delegate as? UIView)?.setNeedsLayout()
```
ephemer, line 138: `setsNeedLayout` should not be the cause of hundreds of re-renders – it deliberately only does layout once per frame.

the reason we do this is because that's what's happens on iOS – removing it would not be correct and would break other things.

I would suggest looking into calling `round` on the dimensions in the CGSize equality check – if we keep landing in this condition it's probably because of some floating point inaccuracies

ephemer (changes requested):
I don't think this is the right way about fixing this issue, see comment inline

Sources/CALayer.swift
ephemer: ok i think i see what you mean though – if we change the bounds twice in a render frame, even if the second `bounds` is equal to what it was at the start of the frame, we will still run layout

ephemer: so instead of comparing `newBounds` to the existing `bounds` (which could have just been set) we kind of need to save the bounds from the previous render cycle. doing this would be easy, but doing it well would take some thought – would you be interested in doing some brainstorming about this?

michaelknoch: sure! brainstorming sounds good @ephemer 

What you said is exactly whats's happening I think, and we really see that the relayouting happens for several frames because `setNeedsLayout` gets set to true over and over again and then it stops at some point (maybe when frames start to get skipped and then the layouting logic is able to catch up? I don't know yet, let's find it out?

Sources/UIView.swift
```diff
@@ -237,6 +237,12 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
     open func layoutSubviews() {
         needsLayout = false
+        
+        if layer.previousBounds != nil && layer.previousBounds != layer.bounds {
+            (layer.superlayer?.delegate as? UIView)?.setNeedsDisplay()
```
ephemer, line 242: Should be setNeedsLayout

ephemer (changes requested):
As discussed on slack this can lead to incorrect layout for a single frame. The question is whether this is an acceptable compromise for now

Sources/UIView.swift
```diff
@@ -237,6 +237,12 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
 
     open func layoutSubviews() {
         needsLayout = false
+        
+        if layer.previousBounds != nil && layer.previousBounds != layer.bounds {
```
ephemer, line 241: Is there no "layouting step" in our render code?
Does layout occur entirely in "user space", meaning it's possible for users to prevent layout on entire subtrees of views? If so, that also doesn't seem quite right, but maybe it's the case in iOS too.

ephemer: I had a look and we do have a "non-userspace" layouting function, which would be a more appropriate place for this, if we went with this option: `Sources/UIView+SDL.swift`

ephemer: But as discussed on Slack, I think a Set containing the updated layers that gets checked and cleared before this layouting step even begins would be the correct place for this

michaelknoch:
closed in favour of https://github.com/flowkey/UIKit-cross-platform/pull/329

PR #329, opened 2021-03-25
do not layout `superview.delegate` from presentation layer

michaelknoch:
**Type of change:** performance improvement

This is a follow up based on https://github.com/flowkey/UIKit-cross-platform/pull/328.

I noticed that the recursive relayouting only happens when wrapping the layout code in an animation block. During this animation in the learnstep we mutate bounds of the presentation on every frame which then triggers layouting for the parent view for each frame as well. I think this matches the ios behaviour.

### DIFF
--- a/Sources/CALayer.swift
+++ b/Sources/CALayer.swift
@@ -133,7 +133,7 @@ open class CALayer {
             guard newBounds != bounds else { return }
             onWillSet(keyPath: .bounds)
 
-            if bounds.size != newBounds.size {
+            if !isPresentationForAnotherLayer && bounds.size != newBounds.size {
                 // It seems weird to access the superview here but it matches the iOS behaviour
                 (self.superlayer?.delegate as? UIView)?.setNeedsLayout()
             }
@@ -226,18 +226,17 @@ open class CALayer {
         return CALayer.defaultAction(forKey: event)
     }
 
-
     /// returns a non animating copy of the layer
     func createPresentation() -> CALayer {
         let copy = CALayer(layer: self)
-        copy.disableAnimations = true
+        copy.isPresentationForAnotherLayer = true
         return copy
     }
 
     internal var _presentation: CALayer?
     open func presentation() -> CALayer? { return _presentation }
 
-    internal var disableAnimations = false
+    internal var isPresentationForAnotherLayer = false
 
     internal var animations = [String: CABasicAnimation]() {
         didSet { onDidSetAnimations(wasEmpty: oldValue.isEmpty) }
--- a/UIKitTests/CALayerTests.swift
+++ b/UIKitTests/CALayerTests.swift
@@ -87,17 +87,15 @@ extension CALayerTests {
         let view = UIView()
         superView.addSubview(view)
         view.layer.bounds.size = CGSize(width: 10, height: 10)
-
         superView.layoutIfNeeded()
         view.layoutIfNeeded()
-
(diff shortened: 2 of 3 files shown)


ephemer (approved):
It would be great if you could add a test that also runs on iOS to verify this behaviour but it seems plausible

UIKitTests/CALayerTests.swift
```diff
@@ -69,4 +69,88 @@ class CALayerTests: XCTestCase {
...
+
+        let superView = TestView()
+        let view = TestView()
+        superView.addSubview(view)
+        view.layer.bounds.size = CGSize(width: 10, height: 10)
+
+        superView.layoutIfNeeded()
+        view.layoutIfNeeded()
+
+        view.layer.bounds.size = CGSize(width: 15, height: 15)
```
michaelknoch, line 94: interestingly it does also relayout the superlayer when mutating `view.layer.bounds.origin`, but I think we should not mimic this behaviour because this also means that we relayout superlayers of UIScrollViews on every scrolled pixel which is not the case on iOS! (see the test below).

There seems to be a different connection between `contentOffset` and `bounds.origin` between ios and our UIKit 

@ephemer

ephemer: I think we'd need to discuss this point on a call – not entirely sure what you mean about the scroll view right now. Is it possible to prevent the infinite layouts on the scrollview by changing that layer's delegate somehow?

I thought we only do a re-layout when the `bounds.size` changes though, so I don't understand how `contentOffset` or `bounds.origin` are related?

Sources/CALayer.swift
```diff
@@ -133,7 +133,7 @@ open class CALayer {
             guard newBounds != bounds else { return }
             onWillSet(keyPath: .bounds)
 
-            if bounds.size != newBounds.size {
+            if !isPresentationForAnotherLayer && bounds.size != newBounds.size {
```
ephemer, line 136: how is this related to `contentOffset` @michaelknoch? That's the bit I don't understand right now

UIKitTests/CALayerTests.swift
michaelknoch: Setting contentOffset directly changes the bounds.origin of the scrollviews layer. On iOS and in our uikit - but I agree, could be more efficient to discuss this in a call. I'll send out an invite for next week.

- Anyways, the pr is ready for being merged. But I would like to share my findings about scrolllview bounds with you

ephemer: @michaelknoch the behaviour we're talking about is only related to `bounds.size`, no? How is that related to bounds.origin / contentOffset?

Sources/CALayer.swift
michaelknoch: on iOS this Check does not exist. They seem to always relayout the superlayer when bounds changes. 

But there is this exception where they do not relayout when setting the content offset of a scrollview.

This surprised me because from our understanding setting the contentOffset directly mutates the bounds.origin.


What does this mean? Initially
I thought we should remove the check and always relayout when bounds change (not only when bounds.size changes) but then we see this massive amount of relayouts when we scroll in our scrollview. Does this make sense @ephemer ? It's pretty hard to explain, it will be way easier to demonstrate.

ephemer: ahhhh ok sorry I didn't get the point about iOS also laying out the superview when `.origin` changes as well

PR #330, opened 2021-04-06
fix multiline text size calculcation

michaelknoch:
Fixes https://app.asana.com/0/1190553484415965/1200141498205813

**Type of change:** bugfix

There is a bug in the text rendering logic, where the frame of a UILabel does not match with the actual surface being rendered by SDL_TTF. This happens when the text renders very close to the right edge of the frame.

I revisited the code in `TTF_RenderUTF8_Blended_Wrapped` and found a minor difference there with what we have in the font renderer.

#### broken:
Screenshot 2021-04-06 at 18 55 47

Screenshot 2021-04-06 at 20 22 12

#### fixed
Screenshot 2021-04-06 at 20 31 02

### DIFF
--- a/Sources/FontRenderer.swift
+++ b/Sources/FontRenderer.swift
@@ -97,17 +97,17 @@ extension FontRenderer {
     }
 
     internal func multilineSize(of text: String, wrapLength: UInt) -> CGSize {
-        return multilineSize(of: text, textLength: Int32(text.count), wrapLength: Int(wrapLength))
+        return multilineSize(of: text, wrapLength: Int(wrapLength))
     }
 
-    private func multilineSize(of text: UnsafePointer<CChar>, textLength: Int32, wrapLength: Int) -> CGSize {
+    private func multilineSize(of text: UnsafePointer<CChar>, wrapLength: Int) -> CGSize {
         guard wrapLength > 0 else { return .zero }
         let lineSpace = 2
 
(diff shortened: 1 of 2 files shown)


Sources/FontRenderer.swift
```diff
@@ -118,7 +118,7 @@ extension FontRenderer {
             var searchIndex =
                 strchr(tok, newLineR) ??
                     strchr(tok, newLineN) ??
-                    end.advanced(by: -1)
```
michaelknoch, line 121: `end.advanced(by: 1)` would also do the trick probably

Sources/FontRenderer.swift
```diff
@@ -97,17 +97,17 @@ extension FontRenderer {
...
-    private func multilineSize(of text: UnsafePointer<CChar>, textLength: Int32, wrapLength: Int) -> CGSize {
+    private func multilineSize(of text: UnsafePointer<CChar>, wrapLength: Int) -> CGSize {
         guard wrapLength > 0 else { return .zero }
         let lineSpace = 2
 
         var textLineHeight: Int32 = 0
 
         var tok = UnsafeMutablePointer(mutating: text)
-        let end = tok + Int(textLength)
+        let end = tok + SDL_strlen(text)
```
michaelknoch, line 110: `SDL_strlen == text.count + 1`

ephemer (approved):

PR #333, opened 2021-07-15
Fix touch

michaelknoch:
Fixes # 

**Type of change:** 

- revert https://github.com/flowkey/UIKit-cross-platform/commit/9a698b9914e62e4623704ee0fcefb14602b88e9c
- fix gesture recognizer cancelling behaviour
- fix responder chain


Sources/UIPanGestureRecognizer.swift
```diff
@@ -63,10 +63,10 @@ open class UIPanGestureRecognizer: UIGestureRecognizer {
     open override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent) {
         super.touchesBegan(touches, with: event)
         guard let trackedTouch = touches.first else { return }
-        state = .began
         self.trackedTouch = trackedTouch
         initialTouchPoint = trackedTouch.location(in: nil)
         touchesMovedTimestamp = trackedTouch.timestamp
+        state = .began
```
michaelknoch, line 69: this is quite important, otherwise `self.trackedTouch` is still `nil` when touchesBegan is executed

Sources/UIScrollView.swift
```diff
@@ -110,8 +110,9 @@ open class UIScrollView: UIView {
         switch panGestureRecognizer.state {
         case .began:
             showScrollIndicators()
-            delegate?.scrollViewWillBeginDragging(self)
             cancelDeceleratingIfNeccessary()
+        case .changed:
+            delegate?.scrollViewWillBeginDragging(self)
```
michaelknoch, line 115: this seems to match iOS behaviour

Sources/UIGestureRecognizer.swift
```diff
@@ -78,24 +85,18 @@ private extension UIGestureRecognizer {
     private func cancelOtherGestureRecognizersThatShouldNotRecognizeSimultaneously() {
         guard let touch = self.trackedTouch else { return }
 
-        let otherRecognizersThatHaveBeganToRecogize = touch.gestureRecognizers.filter {
-            $0 != self && $0.state == .began
+        let otherRecognizers = touch.gestureRecognizers.filter {
+            $0 != self && ($0.state == .began || $0.state == .possible)
```
michaelknoch, line 89: not sure if this internally matches iOS behaviour, but it was needed to achieve the expected UX behaviour. We do now cancel any other recognizer proactively. In our case this means we cancel the sheet pan recognizer when dragging the loop handle

Sources/UITouch.swift
```diff
@@ -53,8 +53,6 @@ public class UITouch {
             action(recognizer)
         }
     }
-
-    internal var hasBeenCancelledByAGestureRecognizer = false
```
michaelknoch, line 57: I don't know why we decided to change the computed to a stored property, maybe the stored prop has advantages I'm currently missing. I started my work by reverting https://github.com/flowkey/UIKit-cross-platform/commit/9a698b9914e62e4623704ee0fcefb14602b88e9c which introduced the buggy behaviour. 

It was already kinda buggy even before this commit, but it was usable

circle.yml
```diff
@@ -1,77 +1,70 @@
...
+            xcode: "12.5.0"
+        steps:
+            - checkout
+            - run: git submodule update --recursive --init
 
-      - run:
-          name: "send coverage report to codecov"
-          command: bash <(curl -s https://codecov.io/bash) -J 'UIKit'
-          environment:
-            - CODECOV_TOKEN: "4b6cd237-4637-42d6-ad29-0d1a6ee59761"
```
michaelknoch, line 25: hoops, prettier crushed the formatting again. I just updated to a newer xcode version and removed uploading the coverage report because it took 5 minutes and sometimes randomly failed. I think we are not even actively using codecov in this project, no?

Sources/UIGestureRecognizer.swift
```diff
@@ -29,17 +29,24 @@ open class UIGestureRecognizer {
 
     public var state: UIGestureRecognizerState = .possible {
         didSet {
-            if state == oldValue { return }
+            if
+                state == oldValue ||
+                // do not mutate state when recognizer was cancelled already.
+                // only allow resetting the state when on `ended`.
```
michaelknoch, line 35:
```suggestion
                // only allow resetting the state on `ended`.
```

Sources/UITouch.swift
ephemer: @michaelknoch can you confirm that the issue #305 was trying to solve is still fixed following the change here? I can see that @janek added tests for that at the time but I don't have a clear picture whether they would have failed before that change. You could find out by reverting that PR except for the tests, and then running the tests (they should fail in that state, and then pass with this PR's changes)

Sources/UIView.swift
```diff
@@ -122,7 +122,7 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
         // }
         didSet {
             didMoveToSuperview()
-            if !(self.next is UIViewController) {
+            if !(self.next is UIViewController) && gestureRecognizers.isEmpty {
```
ephemer, line 125: good catch

Sources/UIView.swift
```diff
@@ -312,6 +312,10 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
     open func addGestureRecognizer(_ recognizer: UIGestureRecognizer) {
         recognizer.view = self
         gestureRecognizers.insert(recognizer)
+
+        if gestureRecognizers.isEmpty {
```
ephemer, line 316: how could this ever happen? wouldn't gestureRecognizers always contain at least one item at this point?

Sources/UIWindow.swift
```diff
@@ -37,8 +37,8 @@ public class UIWindow: UIView {
             currentTouch.view = hitView
             currentTouch.gestureRecognizers = hitView.getRecognizerHierachy()
 
-            currentTouch.runTouchActionOnRecognizerHierachy { $0.touchesBegan(allTouches, with: event) }
             hitView.touchesBegan(allTouches, with: event)
+            currentTouch.runTouchActionOnRecognizerHierachy { $0.touchesBegan(allTouches, with: event) }
```
ephemer, line 41: I'm pretty sure this change is not correct. From my understanding, gesture recognizers need to be able to handle / cancel the touch _before_ touchesBegan gets called

Sources/UIWindow.swift
```diff
@@ -47,16 +47,23 @@ public class UIWindow: UIView {
             }
 
         case .ended:
-            currentTouch.runTouchActionOnRecognizerHierachy { $0.touchesEnded(allTouches, with: event) }
             if !currentTouch.hasBeenCancelledByAGestureRecognizer {
                 hitView.touchesEnded(allTouches, with: event)
             }
 
+            currentTouch.runTouchActionOnRecognizerHierachy { $0.touchesEnded(allTouches, with: event) }
```
ephemer, line 54: this change i'm not so sure about (compared to the touchesBegan one) – it may be ok like this, but my gut feeling is that it was more correct how it was

UIKitTests/Gestures/UIPanGestureRecognizerTests.swift
```diff
@@ -87,7 +87,10 @@ class UIPanGestureRecognizerTests: XCTestCase {
         // expect that state was .cancelled at some point
         wait(for: [cancelledExpectation], timeout: 1)
 
-        // state should transition to .possible after being .cancelled
+        XCTAssert(recognizer.state == .cancelled)
+
+        // state should transition to .possible after ending
+        recognizer.touchesEnded([touch], with: UIEvent())
```
ephemer, line 93: seems reasonable. does this test also run on iOS? (I remember testing touches on iOS was tricky)

michaelknoch: unfortunately not because the test suite is importing UIKit with `@testable`

Sources/UIWindow.swift
michaelknoch: I was able to revert the change without breaking the UX, so it's fine for me to leave it as it was before

Sources/UIView.swift
michaelknoch: good catch, the logic doesn't make much sense. This is what was intended https://github.com/flowkey/UIKit-cross-platform/pull/333/commits/4256b9f28704baeb2278b12e1558d6486841a21b

Sources/UIWindow.swift
michaelknoch: i can keep the old order, we just have to compute `hasBeenCancelledByAGestureRecognizer` before running the recognizer chain

Sources/UITouch.swift
michaelknoch: the tests still pass, and they didn't fail because touchesEnded is still working correctly (not being called when there is a recognizer) - but in our case touchesBegan is the problematic one because it still does get called even when there are recognizers (which hides the draghandles), but since touchesEnded was fixed the dragHandles did not appear again on touch up

not sure if this made any sense. in short: improving touchesEnded behaviour without improving touchesBegan behaviour introduced the bug in our player. this pr brings these behaviours together again. To answer your question: I think it's fine, I also enhanced the existing tests

Sources/UIWindow.swift
```diff
@@ -57,6 +65,16 @@ public class UIWindow: UIView {
...
 
+private extension UITouch {
+    var containsTapRecognizersWhichCancelsTouchesInHitview: Bool {
+        return gestureRecognizers.contains(where: { $0 is UITapGestureRecognizer && $0.state == .began && $0.cancelsTouchesInView })
+    }
+
+    var containsRecognizersWhichCancelsTouchesInHitview: Bool {
+        return gestureRecognizers.contains(where: { ($0.state == .began || $0.state == .changed) && $0.cancelsTouchesInView })
+    }
+}
```
michaelknoch, line 76: we are now checking for .began and .changed, which is correct I think

michaelknoch: because we want tapGestureRecognizers to cancel touchesBegan on hitview (and the hitview responder)

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -236,6 +236,12 @@ public func onNativeTouch(
     pressure: JavaFloat,
     timestampMs: JavaLong
 ) {
+    if env == nil || view == nil {
+        // We're in an invalid state where the JNI has exploded somehow
+        // Passing anything on to UIKit now would probably lead to a crash
+        return
+    }
```
ephemer, line 243: @michaelknoch FYI – I haven't tested this but if you have a device and a working build you could test that it works in the standard case and if so I have a good feeling we might solve most of the UIKit crashes with it

Sources/UIApplication.swift
```diff
@@ -89,7 +89,7 @@ import JNI
 private let maxFrameRenderTimeInSeconds = 1.0 / 60.0
 
 @_cdecl("Java_org_libsdl_app_SDLActivity_nativeProcessEventsAndRender")
-public func nativeProcessEventsAndRender(env: UnsafeMutablePointer<JNIEnv>, view: JavaObject) {
+public func nativeProcessEventsAndRender(env: UnsafeMutablePointer<JNIEnv?>?, view: JavaObject?) {
```
ephemer, line 92: @michaelknoch I also saw some crashes in this function – again these arguments are unused. If this still compiles it may well fix those crashes too

michaelknoch:
I think I addressed your main feedback point from july 2021 in https://github.com/flowkey/UIKit-cross-platform/pull/333/commits/28da989780e81a7ef987dc4efebc2a43a07f8327. As far as I remember you wanted me to decouple `UITouch` and `UIGestureRecognizer`. You think this can go in? Happy to have another session to clarify about details if needed @ephemer .

Sources/UIGestureRecognizer.swift
```diff
@@ -29,17 +29,19 @@ open class UIGestureRecognizer {
...
     public var state: UIGestureRecognizerState = .possible {
         didSet {
-            if state == oldValue { return }
+            if
+                state == oldValue || oldValue == .cancelled && state != .ended
+            {
+                return
+            }
+
             onStateChanged?()
```
ephemer, line 38: why do we return early before calling `onStateChanged`? A code comment would be useful

Sources/UIGestureRecognizer.swift
```diff
@@ -29,17 +29,19 @@ open class UIGestureRecognizer {
...
+
             switch state {
-            case .failed, .cancelled:
-                // touchesCancelled(touches: Set<UITouch>, with: UIEvent)
+            case .recognized, .ended, .failed:
                 state = .possible
```
ephemer, line 42: ok so if we `failed` then we call `onStateChanged` which presumably is aware of that current state, and immediately after that we set the state back to `.possible`?

when does `touchesCancelled` get called?

Sources/UIGestureRecognizer.swift
```diff
@@ -29,17 +29,19 @@ open class UIGestureRecognizer {
...
-            case .recognized, .ended:
-                state = .possible
-            case .changed:
+            case .began, .changed:
```
ephemer, line 43: i don't think it's correct to cancel gesture recognisers on `.began` –  `.began` is an "open" starting state that _might_ lead to something happening but is not yet certain – that's why we used to wait until `.changed` here. I think all of this stuff needs code comments though because it's pretty complex and often unclear why things are the way they are

Sources/UIGestureRecognizer.swift
ephemer: we should cancel the pan recognizer as soon as the loop handle recogniser turns to _`.changed`_, i.e. when it starts actually doing something, not as soon as we click the drag handle (which is what `.began` means – I think!)

Sources/UIGestureRecognizer.swift
```diff
@@ -78,24 +80,18 @@ private extension UIGestureRecognizer {
     private func cancelOtherGestureRecognizersThatShouldNotRecognizeSimultaneously() {
         guard let touch = self.trackedTouch else { return }
 
-        let otherRecognizersThatHaveBeganToRecogize = touch.gestureRecognizers.filter {
-            $0 != self && $0.state == .began
+        let otherRecognizersWhichMayRecognize = touch.gestureRecognizers.filter {
+            $0 != self && ($0.state == .began || $0.state == .possible)
         }
 
-        otherRecognizersThatHaveBeganToRecogize.forEach {
```
ephemer, line 85: the older terminology still makes sense to me logically – it seems correct to me

Sources/UIGestureRecognizer.swift
michaelknoch: but how would that work for TapGestureRecognizers, do they emit `.changed` at all?

Sources/UIPanGestureRecognizer.swift
```diff
@@ -96,6 +96,10 @@ open class UIPanGestureRecognizer: UIGestureRecognizer {
         if state == .changed {
             onAction?()
         }
+
+        if cancelsTouchesInView {
+            trackedTouch.hasBeenCancelledByAGestureRecognizer = true
+        }
```
ephemer, line 102: I do think it makes sense to do this here.

Is it relevant to the logic to have this at the end of the function, or is this something that could (also?) happen earlier? I wonder whether we might have timing bugs doing it at the end of the function like this?

Sources/UIWindow.swift
```diff
@@ -47,8 +50,13 @@ public class UIWindow: UIView {
             }
 
         case .ended:
+            // compute the value before ending the touch on the recognizer hierachy
+            // otherwise `hasBeenCancelledByAGestureRecognizer` will be false because the state was reset already
+            let hasBeenCancelledByAGestureRecognizer = currentTouch.hasBeenCancelledByAGestureRecognizer
```
ephemer, line 55: this comment is very useful, thanks!

UIKitTests/Gestures/TouchHandlingTests.swift
```diff
@@ -40,13 +40,27 @@ class TouchHandlingTests: XCTestCase {
...
+    }
+
+    func testCancelsTouchesInViewWithPanGestureRecognizer() {
         recognizer.cancelsTouchesInView = true
 
         handleTouchDown(CGPoint(x: 10, y: 10))
         handleTouchMove(CGPoint(x: 15, y: 10))
         handleTouchUp(CGPoint(x: 15, y: 10))
 
+        XCTAssertTrue(view.touchesBeganWasCalled)
```
ephemer, line 62: it would be so useful to construct UI Tests out of these for iOS – is that possible?

Sources/UIPanGestureRecognizer.swift
michaelknoch: I think it doesn't matter because `trackedTouch.hasBeenCancelledByAGestureRecognizer` gets only used for the UIView lifecycles which always gets executed after running the actions on the recognizers

Sources/UIGestureRecognizer.swift
ephemer: I actually don't know – I would have expected they should, yes

Sources/UIPanGestureRecognizer.swift
ephemer: Ok awesome. It would be useful if you could add that as a comment (maybe that assumption will be broken in a later code change and it could help debugging)

UIKitTests/Gestures/TouchHandlingTests.swift
michaelknoch: unfortunately no as far as I know, we import UIKit with `@testable import` because we use internal API of UITouch

Sources/UIGestureRecognizer.swift
michaelknoch: thats basically the change I applied to the view lifecycles (recognizer vs view). they don't get called if there is a tap recognizer which has began or a panRecognizer which has moved. This change is not properly reflected in this code here (recognizer vs recognizer). I need to wrap my head around this again.

Sources/UIGestureRecognizer.swift
michaelknoch: I think you are right, and as I said in the comment above this is what we already implemented for the view lifecycles. But it makes sense to also bring this into the gesture logic

I think something like this is correct behaviour
```swift
if 
    $0 === UItapGestureRecognizer && $0.state == .began ||
    $0 === UIPanGestureRecognizer && $0.state == .moved {
        cancel()
}
```

michaelknoch: because AFAIK tap gestures do not "move" at all. I could be wrong though

ephemer: AFAIK there is no such state as `.moved`? Only `.changed`, which is generic for all gesture recognizers. Would it be possible to use `.changed` for all types of recognizers but also ensure the `.changed` state is used for the TapGestureRecognizer? Or does that just not make sense?

Sources/UIGestureRecognizer.swift
ephemer: Ok I am getting the feeling more and more that part of the fix would be ensuring that TapGestureRecognizers get set to the `.changed` state at some point in their lifecycle in order to make things like this work as expected? Would be good to understand what iOS does here

UIKitTests/Gestures/TouchHandlingTests.swift
ephemer: we use the internal API in order to simulate touches programatically, but AFAIK UI Tests can also do that (in a more round-about way). Maybe there's a way of abstracting the simulated touches we do, that on iOS uses the UI Testing APIs and in ours just simulates touches directly?

Sources/UIGestureRecognizer.swift
michaelknoch: this is basically the same as before, we just don't immediately reset the state to `.possible` when cancelled. As far as I can see `touchesCancelled` only gets called from `cancelOtherGestureRecognizersThatShouldNotRecognizeSimultaneously()`

Sources/UIPanGestureRecognizer.swift
```diff
@@ -107,16 +114,10 @@ open class UIPanGestureRecognizer: UIGestureRecognizer {
...
-        super.touchesCancelled(touches, with: event)
-        reset()
-    }
-
     private func reset() {
         trackedTouch = nil
         previousTouchesMovedTimestamp = nil
         touchesMovedTimestamp = nil
         initialTouchPoint = .zero
-        state = .possible
```
ephemer, line 120: i'm a bit concerned about this change. why do we not also reset the state in `reset`? seems surprising to me as a reader

UIKitTests/Gestures/TouchHandlingTests.swift
ephemer: for sure – more wanted to share a general direction we might be able to take

build.gradle
```diff
@@ -25,7 +25,7 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.google.android.exoplayer:exoplayer:2.13.3"
+    implementation "com.google.android.exoplayer:exoplayer:2.13.2"
```
ephemer, line 28: is this change relevant? maybe we can change this in another PR instead

Sources/UIGestureRecognizer.swift
ephemer: i guess this makes sense. would love to better understand what ios does here at some point but doesn't have to be in this PR if it suits our use cases and doesn't do anything too crazy

ephemer (approved):
I have not tested this myself, it would be really great if you could test this on an android device and really try to break it (doing touches with multiple fingers – trying to do the opposite of what we expect users to do) before merging. But code-wise this makes sense to me, thanks for pushing to get this merged @michaelknoch !

ephemer:
Ahh but please address the comment about the `reset()` function before merge as well, and reconsider whether we might want to change the exoplayer version in another PR instead

build.gradle
michaelknoch: this happened because "fix-touch" was our main branch for quite some time. I can cherry-pick this to master before merge to keep the diff clean

PR #334, opened 2021-07-19
fix `xcodebuild` ci step

michaelknoch:
**Type of change:** DX

- update xcode on CI
- remove test which now fails on ios

### DIFF
--- a/UIKitTests/Button/ButtonSetTitleColorForStateTests.swift
+++ b/UIKitTests/Button/ButtonSetTitleColorForStateTests.swift
@@ -17,13 +17,6 @@ class ButtonSetTitleColorForStateTests: XCTestCase {
         button = Button(frame: .zero)
     }
 
-    func testNoTitleColorWithoutTitleForNormalState() {
-        button.setTitleColor(.blue, for: .normal)
-        button.layoutSubviews()
-
-        XCTAssertNotEqual(button.titleLabel?.textColor, .blue)
-    }
-
     func testExistingTitleColorWithTitleForNormalState() {
         button.setTitleColor(.blue, for: .normal)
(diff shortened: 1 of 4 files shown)


UIKitTests/Button/ButtonSetTitleColorForStateTests.swift
```diff
@@ -17,13 +17,6 @@ class ButtonSetTitleColorForStateTests: XCTestCase {
         button = Button(frame: .zero)
     }
 
-    func testNoTitleColorWithoutTitleForNormalState() {
-        button.setTitleColor(.blue, for: .normal)
-        button.layoutSubviews()
-
-        XCTAssertNotEqual(button.titleLabel?.textColor, .blue)
-    }
```
michaelknoch, line 25: this test does now fail when running from the ios test target. I don't fully get what it does to be honest. @ephemer do you think it's worth fixing and keeping this test?

ephemer: really doesn't matter, kill it! I doubt Apple has tests for this internally, I'd say it was just some random undocumented behaviour that we were testing here

ephemer (approved):

UIKitTests/Button/ButtonSetTitleColorForStateTests.swift
ephemer: to be honest it would make more sense to be to turn the test around – to XCTAssertEqual rather than NotEqual. I assume that's what the latest iOS SDK is doing, which makes sense to me. But we'd also need to change the behaviour of what we're doing internally then and I'm not sure it's worth it for a strange edge case

PR #339, opened 2021-09-29
Fix build for macOS with XCode 13

rikner:
Fixes # 

**Type of change:** 

Got some compiler errors when building for macOS with XCode13. For some reason importing the whole Foundation module instead of just the required functions fixes the issue.

Screenshot 2021-09-29 at 13 34 05
Screenshot 2021-09-29 at 13 14 31

### DIFF
--- a/Sources/CAMediaTimingFunction.swift
+++ b/Sources/CAMediaTimingFunction.swift
@@ -6,9 +6,7 @@
 //  Copyright © 2017 flowkey. All rights reserved.
 //
 
-#if os(Android) || os(Linux)
-import func Foundation.pow
-#endif
+import Foundation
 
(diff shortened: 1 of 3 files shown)


mouselangelo (approved):
👍

ephemer (approved):
This works – we are importing all of Foundation in a bunch of other files, so the main reason to restrict what is being imported in these ones is to try to speed up code completion and for "clean code" purposes. In practice the difference isn't huge so your solution looks good to me @rikner 

FYI this issue is very likely due to https://github.com/apple/swift-evolution/blob/main/proposals/0307-allow-interchangeable-use-of-double-cgfloat-types.md

PR #343, opened 2021-11-10
Call presented UIViewController lifecycle methods from UINavigationController

### LINKED ISSUE #341: UINavigationController does not call `viewWillAppear`
## Current behavior
When using a UINavigationController as the root view controller, you can see the Android-style navigation bar at the top but the content is completely black and doesn't seem to show the root view controller that I set inside the UINavigationController.

## Expected behavior
Seeing a normal UINavigationController but with the root view controller actually being shown instead of having a completely black screen below the navigation bar.

## Steps to reproduce
1. Use the sample project and set the root view controller of the window inside AppDelegate.swift to a UINavigationController that has ViewController as its root view controller.
2. Run it on any supported OS (got the same weird results with Android, macOS Monterey and iOS 15.1).

## Screenshots

Screen Shot 2021-11-07 at 18 39 49

## Environment
- Device: Xiaomi A2
- OS: Android Q
- UIKit Version: Latest commit (f8cb41875a3d7a6fa726170e48c68cf0498bd5ac)

ephemer:
Fixes #341 

**Type of change:** Bug fix

As reported in #341, UINavigationController does not work as expected. This is because `viewWillAppear` (etc.) are never called there. I have added a call to `viewWillAppear` and `viewDidAppear`, but there are probably other lifecycle methods that are currently not being called.

### DIFF
--- a/Sources/UINavigationController.swift
+++ b/Sources/UINavigationController.swift
@@ -65,7 +65,7 @@ open class UINavigationController: UIViewController {
     }
 
     open override func dismiss(animated: Bool, completion: (() -> Void)? = nil) {
-        if self === UIApplication.shared.keyWindow?.rootViewController {
+        guard self !== UIApplication.shared.keyWindow?.rootViewController else {
             completion?()
             return
         }
--- a/Sources/UINavigationController.swift
+++ b/Sources/UINavigationController.swift
@@ -65,6 +65,11 @@ open class UINavigationController: UIViewController {
     }
 
     open override func dismiss(animated: Bool, completion: (() -> Void)? = nil) {
+        if self === UIApplication.shared.keyWindow?.rootViewController {
+            completion?()
+            return
+        }
+
         let topOfStack = viewControllers.last
         topOfStack?.viewWillDisappear(animated)
 
--- a/Sources/UINavigationController.swift
+++ b/Sources/UINavigationController.swift
@@ -33,6 +33,7 @@ open class UINavigationController: UIViewController {
 
     private func updateUIFromViewControllerStack(animated: Bool) {
         guard _view != nil else { return }
+        if presentedViewController === viewControllers.last { return }
 
         transitionView.subviews.forEach { $0.removeFromSuperview() }
         presentedViewController = viewControllers.last
@@ -40,7 +41,9 @@ open class UINavigationController: UIViewController {
         if let viewOnTopOfStack = viewControllers.last?.view {
             // TODO: Animate here
             viewOnTopOfStack.frame = transitionView.bounds
+            presentedViewController?.viewWillAppear(animated)
             transitionView.addSubview(viewOnTopOfStack)
+            presentedViewController?.viewDidAppear(animated)
         }
     }
 
--- a/Sources/UINavigationController.swift
+++ b/Sources/UINavigationController.swift
@@ -65,7 +65,7 @@ open class UINavigationController: UIViewController {
     }
 
     open override func dismiss(animated: Bool, completion: (() -> Void)? = nil) {
-        guard self !== UIApplication.shared.keyWindow?.rootViewController else {
+        guard self !== UIApplication.shared?.keyWindow?.rootViewController else {
             completion?()
             return
         }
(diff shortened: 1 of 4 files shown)


rikner:
Hey @ephemer ! 
I reproduced the issue this way (AppDelegate.swift):
```
let viewController = ViewController()
let navigationController = UINavigationController(rootViewController: viewController)
window?.rootViewController = navigationController
```
The PR seems to fix the issue, but I realized that the BackButton does not seem to work. Tapping it has no effect. Using the hardware button (or actually the screen gesture in my case) works.

https://user-images.githubusercontent.com/2410252/141165039-abe0a397-7ffe-4aea-afa6-3855764a983f.mp4

ephemer:
Hey @rikner, thanks a lot for testing this!

I disabled the back button specifically in this PR because it should not do anything when the `UINavigationController` is at root position. If the back button still "works" we need to fix that too :)

rikner:
> Hey @rikner, thanks a lot for testing this!
> 
> I disabled the back button specifically in this PR because it should not do anything when the `UINavigationController` is at root position. If we dismiss the rootViewController the app will be "dead" forever. If the hardware back button still "works" we need to fix that too :)

Ah ok I see. Apparently I have no idea what the expected behaviour is. Are you looking into disabling the Hardware Button?

ephemer:
@rikner the hardware back button should work in general, but the "back button" on a UINavigationController should not do anything if there is only one view on the stack _and_ that UINavigationController is the application's `rootViewController`

rikner:
> @rikner the hardware back button should work in general, but the "back button" on a UINavigationController should not do anything if there is only one view on the navigation stack _and_ that UINavigationController is the application's `rootViewController`

Ok, yeah I get that. But then shouldn't we make sure that in this special case the hardware back button does not send the App into the void like seen in the video? Or would that be an issue for a seperate PR?

ephemer:
Yep:

> If the hardware back button still "works" we need to fix that too :)

Sources/UINavigationController.swift
```diff
@@ -62,6 +65,11 @@ open class UINavigationController: UIViewController {
     }
 
     open override func dismiss(animated: Bool, completion: (() -> Void)? = nil) {
+        if self === UIApplication.shared.keyWindow?.rootViewController {
+            completion?()
+            return
+        }
```
ephemer, line 71: @rikner this should fix the issue you were seeing

samples/getting-started/DemoApp/AppDelegate.swift
```diff
@@ -19,10 +19,9 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
     ) -> Bool {
 
         window = UIWindow()
-        window?.rootViewController = ViewController()
+        window?.rootViewController = UINavigationController(rootViewController: ViewController())
```
ephemer, line 22: @michaelknoch any objections to this?

Sources/UINavigationController.swift
ephemer: I tested this in the Mac App (Esc has the same effect as the physical back button on Android) and it works as expected (as in, it doesn't do anything) 👍🏼

michaelknoch (approved):
looks good to me!

ephemer:
@rikner @michaelknoch would appreciate a(nother) review on this

samples/getting-started/DemoApp/AppDelegate.swift
michaelknoch: nope, makes sense I think

michaelknoch:
oops, CI doesn't like the change @ephemer 😄 
https://app.circleci.com/pipelines/github/flowkey/UIKit-cross-platform/151/workflows/202a524e-10dc-4ce7-9862-0da1201f167d/jobs/1453

ephemer:
@michaelknoch `UIViewControllerTests.testMemoryLeakInNavigationController` seems very suspicious ;) I will check it out thanks

ephemer:
ah, it's because `navigationController!.dismiss(animated: false)` will now only work if it's not the root view controller, which is desired behaviour. I will fix it

Sources/UINavigationController.swift
rikner: yes, can confirm 👍

rikner (approved):

PR #344, opened 2021-11-10
Reduce dependency on Foundation

ephemer:
**Type of change:** DX

`Foundation` is huge in file size and complexity. It is difficult to build, tedious to update, and relying on it is blocking us from moving to new platforms (like the browser). As an open source project swift-corelibs-foundation has not gained a huge amount of traction and appears to be only implementing the bare minimum. In general, it seems like the Swift community is looking to replace it with lighter more specific modules. We would also do better to remove it and implement what we need from it in other ways wherever possible.

This PR makes it clear that we are really only relying on a few key parts of `Foundation` in UIKit-Cross-Platform:

- `URL` (in `AVURLAsset`, which can be split out and implemented differently)
- `Data`
- `NSAttributedString` (which also requires `NSRange` for our purposes, but can probably be removed entirely)
- `Notification(Center)`
- `Date` (only to get the current timestamp – can be implemented differently)

Some of these dependencies will be easier to to remove than others, and that work should happen in a separate PR. For now, we just want to make explicit what we are _actually_ using and remove some trivial dependencies directly.

### DIFF
--- a/Sources/CATransform3D.swift
+++ b/Sources/CATransform3D.swift
@@ -6,8 +6,6 @@
 //  Copyright © 2018 flowkey. All rights reserved.
 //
 
-import Foundation
-
 public struct CATransform3D {
     public init(
         m11: Float, m12: Float, m13: Float, m14: Float,
@@ -49,12 +47,16 @@ extension CATransform3D {
 
 internal extension CATransform3D {
     func transformingVector(x: CGFloat, y: CGFloat, z: CGFloat) -> (x: CGFloat, y: CGFloat, z: CGFloat) {
-        let newX = CGFloat(m11) * x + CGFloat(m21) * y + CGFloat(m31) * z + CGFloat(m41)
-        let newY = CGFloat(m12) * x + CGFloat(m22) * y + CGFloat(m32) * z + CGFloat(m42)
-        let newZ = CGFloat(m13) * x + CGFloat(m23) * y + CGFloat(m33) * z + CGFloat(m43)
-        let newW = CGFloat(m14) * x + CGFloat(m24) * y + CGFloat(m34) * z + CGFloat(m44)
+        let newX = Double(m11) * Double(x) + Double(m21) * Double(y) + Double(m31) * Double(z) + Double(m41)
+        let newY = Double(m12) * Double(x) + Double(m22) * Double(y) + Double(m32) * Double(z) + Double(m42)
+        let newZ = Double(m13) * Double(x) + Double(m23) * Double(y) + Double(m33) * Double(z) + Double(m43)
+        let newW = Double(m14) * Double(x) + Double(m24) * Double(y) + Double(m34) * Double(z) + Double(m44)
 
-        return (x: newX / newW, y: newY / newW, z: newZ / newW)
+        return (
+            x: CGFloat(newX / newW),
+            y: CGFloat(newY / newW),
+            z: CGFloat(newZ / newW)
+        )
     }
 }
 
--- a/Sources/CGImage.swift
+++ b/Sources/CGImage.swift
@@ -8,7 +8,7 @@
 
 import SDL
 import SDL_gpu
-import Foundation
+import struct Foundation.Data
 
 public class CGImage {
     /// Be careful using this pointer e.g. for another CGImage instance.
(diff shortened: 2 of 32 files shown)


Sources/CATransform3D.swift
```diff
@@ -49,12 +47,16 @@ extension CATransform3D {
...
+        let newX = Double(m11) * Double(x) + Double(m21) * Double(y) + Double(m31) * Double(z) + Double(m41)
+        let newY = Double(m12) * Double(x) + Double(m22) * Double(y) + Double(m32) * Double(z) + Double(m42)
+        let newZ = Double(m13) * Double(x) + Double(m23) * Double(y) + Double(m33) * Double(z) + Double(m43)
+        let newW = Double(m14) * Double(x) + Double(m24) * Double(y) + Double(m34) * Double(z) + Double(m44)
+
+        return (
+            x: CGFloat(newX / newW),
+            y: CGFloat(newY / newW),
+            z: CGFloat(newZ / newW)
+        )
```
rikner, line 59: Is there any advantage of doing it like this? seems a bit more verbose with all those `Double(...)`

Sources/CGImage.swift
```diff
@@ -50,10 +50,13 @@ public class CGImage {
 
     internal convenience init?(_ sourceData: Data) {
         var data = sourceData
-        let dataCount = Int32(data.count)
 
-        guard let gpuImagePtr = data.withUnsafeMutableBytes({ (ptr: UnsafeMutablePointer<Int8>) -> UnsafeMutablePointer<GPU_Image>? in
-            let rw = SDL_RWFromMem(ptr, dataCount)
+        guard let gpuImagePtr = data.withUnsafeMutableBytes({ buffer -> UnsafeMutablePointer<GPU_Image>? in
```
rikner, line 54: hmm interesting...buffer is now an `UnsafeBufferPointer` right? So previously we used it as an `UnsafeMutablePointer<Int8>` which was not so correct?

Sources/UINavigationBarAndroid.swift
```diff
@@ -27,7 +27,12 @@ open class UINavigationBarAndroid: UINavigationBar {
                 originalLeftButtonOnPress?()
             } else {
                 // We are the last item, or there are no items. Dismiss!
-                (self?.next as? UINavigationController)?.dismiss(animated: true)
+                if
+                    let navigationController = self?.next as? UINavigationController,
+                    navigationController !== UIApplication.shared.keyWindow?.rootViewController
+                {
+                    navigationController.dismiss(animated: true)
+                }
```
rikner, line 35: not really part of this PR.

Sources/UINavigationController.swift
```diff
@@ -33,14 +33,17 @@ open class UINavigationController: UIViewController {
...
 
         transitionView.subviews.forEach { $0.removeFromSuperview() }
         presentedViewController = viewControllers.last
 
         if let viewOnTopOfStack = viewControllers.last?.view {
             // TODO: Animate here
             viewOnTopOfStack.frame = transitionView.bounds
+            presentedViewController?.viewWillAppear(animated)
             transitionView.addSubview(viewOnTopOfStack)
+            presentedViewController?.viewDidAppear(animated)
```
rikner, line 46: also these changes aren't related to this PRs intention.

rikner (approved):
👍

Sources/UINavigationController.swift
ephemer: sorry the PR base was wrong. based this on the previous PR because I plan to merge it first

Sources/CATransform3D.swift
ephemer: Due to https://github.com/apple/swift-evolution/blob/main/proposals/0307-allow-interchangeable-use-of-double-cgfloat-types.md we get all sorts of type errors compiling this (and other) code. Ideally I'd like to replace CGFloat with Double in general, then we could remove all of them. For now I'd like to do all the calculations using Doubles and return CGFloats (which may be lower precision in the current implementation), which will provide a more accurate answer

Sources/CGImage.swift
ephemer: The old API is deprecated – the new code essentially does the same thing though. I just took the opportunity to get rid of some build warnings while cleaning up the other bits

ephemer:
FYI @rikner @michaelknoch I am going to merge this. The tests are passing and this PR should not introduce any behavioural change 🙏🏼

PR #345, opened 2021-11-26
maybe fix crashes in touch handling

michaelknoch:
**Type of change:** UX

I cherry picked the changes from the `fix-touch` branch. I'm not entirely sure if they improved the situation, but I thought we have slightly less crashes, no?


rikner (approved):
I think we still saw the crashes while we used the `fix-touch` branch in production. But I guess it does no harm applying these commits anyway.

ephemer (approved):
Honestly I don't think this made much difference after all but it's not going to be harmful to merge it

PR #346, opened 2021-11-26
fix surfaceCreated override (#342)

michaelknoch:
**Type of change:** fix build

also picked from `fix-touch` pr

### DIFF
--- a/src/main/java/org/libsdl/app/SDLActivity.kt
+++ b/src/main/java/org/libsdl/app/SDLActivity.kt
@@ -247,7 +247,7 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
         mSurface.setOnTouchListener(this)
     }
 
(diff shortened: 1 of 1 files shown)


rikner (approved):

src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -247,7 +247,7 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
         mSurface.setOnTouchListener(this)
     }
 
-    override fun surfaceCreated(holder: SurfaceHolder?) {
```
michaelknoch, line 250: why did this change actually, is this because of an SDK update? we should probably change the samples app as well then @rikner

rikner: Yeah, pretty sure it was because of the SDK update 👍

PR #340, opened 2021-09-30
Update ExoPlayer

rikner:
**Type of change:** maintenance

## Motivation (current vs expected behavior)
Update ExoPlayer to the [earliest version available on Maven](https://mvnrepository.com/artifact/com.google.android.exoplayer/exoplayer) and kick jcenter().
Replace all of the deprecated API calls.

## Please check if the PR fulfills these requirements
- [x] Self-review: I am confident this is the simplest and clearest way to achieve the expected behaviour
- [x] There are no dependencies on other PRs or I have linked dependencies through Zenhub
- [x] The commit messages are clean and understandable
- [ ] Tests for the changes have been added (for bug fixes / features)

PR #347, opened 2021-11-29
fix android build of sample app

michaelknoch:
**Type of change:** DX

- update to sdk 30
- remove arm32 build
- call cmake from gradle instead of calling cmake from cmake


samples/getting-started/android/app/build.gradle
```diff
@@ -35,6 +30,29 @@ android {
...
+preBuild.dependsOn buildSwiftArm64
+preBuild.dependsOn buildSwiftIntel64
+
+clean.dependsOn cleanSwift
+
+def getSwiftBuildCommand(abi) {
+    def cmakeBuildType = "RelWithDebInfo"
+    def buildScript = "../../../../swift-android-toolchain/swift-build.sh"
+    return "CMAKE_BUILD_TYPE=$cmakeBuildType ANDROID_ABI=$abi LIBRARY_OUTPUT_DIRECTORY='${projectDir}/src/main/jniLibs/${abi}' $buildScript ../../"
+}
```
rikner, line 54: calling cmake from gradle instead inside the other cmake seems easier to grasp 👍  is there any other advantage?

samples/getting-started/android/build.gradle
```diff
@@ -4,6 +4,7 @@ buildscript {
     ext.kotlin_version = '1.5.31'
     repositories {
         google()
+        jcenter()
```
rikner, line 7: I removed this `jcenter()` in [this PR](https://github.com/flowkey/UIKit-cross-platform/pull/340/files#diff-2e072c2476eea21c7c03216f034ba0827afea76c24e55b05f2f37693592b7f9bL7) a while ago. What dependency are does depend on it and can we instead use maven?

rikner (approved):

samples/getting-started/android/build.gradle
michaelknoch: fair point, It was failing without jcenter but only because that specific version of the build tools didn't exist in the maven or google repo.

this commit fixes the build without jcenter()
https://github.com/flowkey/UIKit-cross-platform/pull/347/commits/d4234be2b9ebefaa59f478ce99a376de2d753c88

samples/getting-started/android/app/build.gradle
michaelknoch: we mainly changed this because with the "calling cmake from cmake" approach we sometimes were missing the libs in the final apk for some weird reason. I think this is a bug in the gradle tools somewhere. It never happened with the new approach because explicitly build the libs on `preBuild`

michaelknoch: another benefit is that we now see all swift build logs and errors. with the old approach it was usually hidden behind "building cmake..."

samples/getting-started/android/app/build.gradle
```diff
@@ -3,11 +3,11 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 30
```
ephemer, line 6: @rikner do we need to do anything regarding tagged pointers here too or is that only from v31+?

rikner: ah sure right! we must disable tagged pointers in the sample apps manifest, see https://github.com/flowkey/NativePlayer/pull/898/files#diff-dadb150a88f63da91485cd57b79259d15a19bb20782f0f0bb3795237b926171bR14
otherwise we would risk crashes.

ephemer: ok, would you be able to make a PR for that? (not urgent)

PR #351, opened 2021-12-08
Implement basic Switch

rikner:
Fixes # 

**Type of change:** 

This is a minimal implementation for a UISwitch-like component

Resources: https://material.io/develop/android/components/switches

Jan-07-2022 15-28-50

now with thumb having a shadow:
switch

### DIFF
--- a/Sources/Switch.swift
+++ b/Sources/Switch.swift
@@ -2,7 +2,10 @@ private let frameSize = CGSize(width: 51, height: 31)
 private let thumpSize = frameSize.height - 3
 private let animationTime = 0.25
 
-open class UISwitch: UIControl {
+// Note: we deliberately don't wrap UISwitch.
+// This allows us to have a somewhat custom API free of objc selectors etc.
+
+open class Switch: UIControl {
 
     public let tapGestureRecognizer = UITapGestureRecognizer()
 
--- a/Sources/Switch.swift
+++ b/Sources/Switch.swift
@@ -15,7 +15,7 @@ private let offTrackColor = CGColor(red: 225 / 255, green: 225 / 255 , blue: 225
 open class Switch: UIControl {
     private let tapGestureRecognizer = UITapGestureRecognizer()
     
-    private var thumb = UIView()
+    private var thumb = SwitchThumb()
 
     public var onTintColor: UIColor? = defaultOnTintColor
     public var thumbTintColor: UIColor? = defaultThumbColor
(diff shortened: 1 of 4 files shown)


Sources/Switch.swift
```diff
@@ -0,0 +1,50 @@
+private let frameSize = CGSize(width: 51, height: 21)
+private let thumbSize = CGFloat(31.0)
+private let animationTime = 0.25
+
+private let defaultOnTintColor = CGColor(red: 0.69, green: 0.5, blue: 0.96, alpha: 1)
+private let defaultThumbColor = CGColor(red: 0.30, green: 0, blue: 0.91, alpha: 1)
```
rikner, line 6: this could be basically any colours

michaelknoch:
it seems that the switch is invisible on first render and has to be clicked once to make it visible @rikner 
https://beta.record.wtf/play?id=1641471935790_47345

rikner:
> it seems that the switch is invisible on first render and has to be clicked once to make it visible @rikner https://beta.record.wtf/play?id=1641471935790_47345

thx for the hint. fixed it 👍

michaelknoch (approved):

Sources/Switch.swift
```diff
@@ -0,0 +1,96 @@
+private let frameSize = CGSize(width: 51, height: 21)
+private let thumbSize = CGFloat(31.0)
+private let animationTime = 0.25
+
+private let defaultOnTintColor = CGColor(red: 169 / 255, green: 218 / 255 , blue: 214 / 255, alpha: 1)
+private let defaultThumbColor = CGColor(red: 14 / 255, green: 136 / 255 , blue: 122 / 255, alpha: 1)
+
+private let offThumbColor = CGColor(red: 176 / 255, green: 176 / 255 , blue: 176 / 255, alpha: 1)
+private let offTrackColor = CGColor(red: 225 / 255, green: 225 / 255 , blue: 225 / 255, alpha: 1)
```
michaelknoch, line 9: nit: personally I would move these private styling vars to the bottom of the file

michaelknoch (approved):

mouselangelo:
Hey @rikner - sorry I never got around to reviewing this! I had reviewed the code and wanted to test this locally but just didn't get around to doing that! The implementation looks great though. Thanks 🙏 

I am wondering if adding this to the `samples/getting-started` app, like we did for alerts UI would have made it easier to test. I also like the idea of being able to preview all the components and test them.

I'd even suggest implementing a new `samples/storybook` or `samples/showcase` project - to keep `getting-started` very simple. We could implement something similar to [storybook](https://next--storybookjs.netlify.app/official-storybook), but purely using UIKit. We use storybook in our harmony design system and it's very useful to browse through the various core styles and components.

I'd be happy to support with this. @michaelknoch @ephemer - what do you guys think?

PR #352, opened 2021-12-09
Implement UIAlertControllerStyle.alert

rikner:
**Type of change:** feature

So far we only handled `.actionSheet` style for UIAlertControllers. This PR adds code to handle the `.alert` style.

alert style:
Bildschirmfoto 2021-12-09 um 19 09 53

actionSheet for reference:
Bildschirmfoto 2021-12-09 um 19 09 56

demo:
demo

### DIFF
--- a/Sources/UIAlertController.swift
+++ b/Sources/UIAlertController.swift
@@ -23,7 +23,6 @@ public class UIAlertController: UIViewController {
 
     public init(title: String?, message: String?, preferredStyle: UIAlertControllerStyle) {
         self.message = message
-        assert(message == nil, "We haven't implemented `message` yet")
         self.preferredStyle = preferredStyle
         super.init(nibName: nil, bundle: nil)
         self.title = title
--- a/Sources/UIAlertController.swift
+++ b/Sources/UIAlertController.swift
@@ -10,7 +10,6 @@ import func Foundation.round
 
 public enum UIAlertControllerStyle {
     case actionSheet
-    case popover
     case alert
 }
 
--- a/Sources/UIAlertControllerView.swift
+++ b/Sources/UIAlertControllerView.swift
@@ -40,6 +40,8 @@ class UIAlertControllerView: UIView {
         self.style = style
         header.text = title
         header.font = UIFont.boldSystemFont(ofSize: 20)
+        header.numberOfLines = 0
+        header.layer.contentsGravity = .top
 
         if let message = message {
             text = UILabel(frame: .zero)
@@ -72,8 +74,6 @@ class UIAlertControllerView: UIView {
         backgroundColor = .white
         layer.cornerRadius = 3
 
-        header.numberOfLines = 0
-        header.layer.contentsGravity = .top
         addSubview(header)
         text.map { addSubview($0) }
         buttons.forEach { addSubview($0) }
@@ -93,16 +93,12 @@ class UIAlertControllerView: UIView {
         header.frame.origin = CGPoint(x: horizontalPadding, y: verticalPadding)
         header.bounds.width = subviewWidth
         header.sizeToFit()
+        header.textAlignment = .left // sizeToFit() seems to change textAlignment
 
-        // XXX: sizeToFit() seems to change the alignment, so we're explicitly reassigning it here
-        header.textAlignment = .left
-
-        if let text = text {
-            text.frame.origin.x = horizontalPadding
-            text.frame.width = subviewWidth 
-            text.sizeToFit()
-            text.frame.minY = header.frame.maxY + verticalPadding
-        }
+        text?.frame.origin.x = horizontalPadding
+        text?.frame.width = subviewWidth
+        text?.sizeToFit()
+        text?.frame.minY = header.frame.maxY + verticalPadding
 
         if style == .alert {
             let totalButtonsWidth = buttons.reduce(0, { $0 + $1.frame.width }) + horizontalGap * CGFloat(buttons.count - 1)
--- a/Sources/UIAlertControllerView.swift
+++ b/Sources/UIAlertControllerView.swift
@@ -82,6 +82,8 @@ class UIAlertControllerView: UIView {
             }
             return min(maxWidth, max(largestSubviewWidth, minWidth))
         }()
+        
+        subviews.forEach { $0.clipsToBounds = true }
     }
 
     override func layoutSubviews() {
@@ -134,7 +136,6 @@ class UIAlertControllerView: UIView {
     private func layoutButtonsHorizontally() {
         buttons.enumerated().forEach({ index, button in
             button.sizeToFit()
-            button.clipsToBounds = true
             button.frame.width = max(button.frame.width, 75)
             button.contentHorizontalAlignment = .center
             button.frame.minY = CGFloat(text?.frame.maxY ?? header.frame.maxY) + verticalPadding
--- a/Sources/UIAlertControllerView.swift
+++ b/Sources/UIAlertControllerView.swift
@@ -51,7 +51,6 @@ class UIAlertControllerView: UIView {
         layer.cornerRadius = 3
 
         header.sizeToFit()
-        header.frame.height += verticalPadding
         header.layer.contentsGravity = .top
         addSubview(header)
         addSubview(text)
@@ -63,8 +62,6 @@ class UIAlertControllerView: UIView {
         switch style {
         case .actionSheet: layoutAsActionSheet()
         case .alert: layoutAsAlert()
-        default: assertionFailure(
-            "The UIAlertControllerStyle style, \(style), is not implemented yet!")
         }
     }
 
@@ -75,48 +72,55 @@ class UIAlertControllerView: UIView {
             let previousElement = (index > 0) ? buttons[index - 1] : header
             button.sizeToFit()
             button.contentHorizontalAlignment = .left
-            button.bounds.size.height += verticalPadding
+            button.frame.size.height += verticalPadding
             button.frame.size.width = max(bounds.size.width, button.frame.size.width)
-            button.frame.origin = CGPoint(x: 0, y: previousElement.frame.maxY)
+            button.frame.origin.y = (previousElement == header)
+                ? previousElement.frame.maxY + verticalPadding
+                : previousElement.frame.maxY
+            button.titleLabelOriginX = horizontalPadding
         })
     }
 
     private func layoutAsAlert() {
         header.frame.origin = CGPoint(x: horizontalPadding, y: verticalPadding)
         text.frame.origin.x = horizontalPadding
-        text.bounds.width = UIScreen.main.bounds.width * 0.5
+        text.frame.width = UIScreen.main.bounds.width * 0.5
         text.sizeToFit()
-        text.frame.minY = header.frame.maxY
+        text.frame.minY = header.frame.maxY + verticalPadding
         buttons.enumerated().forEach({ (arg) in
             let (index, button) = arg
-
             button.sizeToFit()
-            button.contentHorizontalAlignment = .left
-            button.frame.size.width = max(bounds.size.width, button.frame.size.width)
-            
-            let previousElement = (index > 0) ? buttons[index - 1] : text
-            if previousElement == text {
-                button.frame.origin.y = previousElement.frame.maxY + verticalPadding
+            button.frame.width = max(button.frame.width, 75)
+            button.contentHorizontalAlignment = .center
+            button.frame.minY = text.frame.maxY + verticalPadding
+            if button == buttons.first {
+                button.frame.origin.x = bounds.maxX - button.frame.width - verticalPadding
             } else {
-                button.frame.origin.y = previousElement.frame.maxY
-            }
-            
-            if button == buttons.last! {
-                button.bounds.size.height += verticalPadding
+                let previousElement = buttons[index - 1]
+                button.frame.origin.x = previousElement.frame.minX - button.frame.width
             }
         })
     }
 
     override func sizeThatFits(_ size: CGSize) -> CGSize {
         layoutIfNeeded()
-
-        let widestElement = ([text] + buttons).max { a, b in a.frame.width < b.frame.width }
-        let elementWidth = widestElement?.frame.width ?? header.frame.width
-        let elementHeight = header.frame.height + text.frame.height + buttons.reduce(0, { $0 + $1.bounds.height })
-
+        
+        let width: CGFloat
+        let height: CGFloat
+        
+        switch style {
+        case .actionSheet:
+            width = ([text] + buttons).map { $0.frame.width }.max()!
+            height = buttons.last?.frame.maxY ?? header.frame.height
+        case .alert:
+            let buttonsWidth = buttons.reduce(0, { $0 + $1.frame.width })
+            width = ([header.frame.width, text.frame.width, buttonsWidth]).max()!
+            height = buttons.first?.frame.maxY ?? text.frame.height
+        }
+        
         return CGSize(
-            width: max(minWidth, elementWidth) + horizontalPadding, // on right
-            height: elementHeight + verticalPadding * 2 // at bottom
+            width: max(minWidth, width) + horizontalPadding,
+            height: height + verticalPadding
         )
     }
 }
@@ -129,9 +133,12 @@ class UIAlertControllerButton: Button {
         }
     }
 
+    var titleLabelOriginX: CGFloat?
+    
     override func layoutSubviews() {
         super.layoutSubviews()
-
-        titleLabel?.frame.origin.x = horizontalPadding
+        if let titleLabelOriginX = titleLabelOriginX {
+            titleLabel?.frame.origin.x = titleLabelOriginX
+        }
     }
 }
--- a/Sources/UIAlertControllerView.swift
+++ b/Sources/UIAlertControllerView.swift
@@ -39,14 +39,13 @@ class UIAlertControllerView: UIView {
     {
         self.style = style
         header.text = title
-        header.font = UIFont.systemFont(ofSize: 20)
+        header.font = UIFont.boldSystemFont(ofSize: 20)
 
         if let message = message {
             text = UILabel(frame: .zero)
             text?.numberOfLines = 0
             text?.text = message
-            text?.font = UIFont.systemFont(ofSize: 16, weight: .light)
-            text?.textColor = UIColor(hex: 0x3C3C43)
+            text?.font = UIFont.systemFont(ofSize: 16)
         } else {
             text = nil
         }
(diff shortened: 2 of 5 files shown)


Sources/UIAlertController.swift
```diff
@@ -10,7 +10,6 @@ import func Foundation.round
 
 public enum UIAlertControllerStyle {
     case actionSheet
-    case popover
```
rikner, line 13: No idea why we had this. It actually does not exist on iOS.
https://developer.apple.com/documentation/uikit/uialertcontroller/style

samples/getting-started/DemoApp/ViewController.swift
```diff
@@ -11,6 +11,11 @@ import UIKit
 class ViewController: UIViewController {
 
     let label = UILabel()
+    let buttonForAlert = Button()
+    let buttonForAction = Button()
+    
+    let alertController = UIAlertController(title: "Alert Message", message: alertMessageShortest, preferredStyle: .alert)
+    let actionsController = UIAlertController(title: "Actions", message: nil, preferredStyle: .actionSheet)
```
rikner, line 18: Thought it might be nice to add this to the DemoApp. But we could also revert all these changes here.

Sources/UIAlertControllerView.swift
```diff
@@ -12,6 +12,7 @@ private let minWidth: CGFloat = 300
 
 class UIAlertControllerView: UIView {
     let header = UILabel(frame: .zero)
+    let text = UITextView(frame: .zero)
```
mouselangelo, line 15: We won't need this view if we have no `message`.
It might be better to declare this as an optional variable here and init it only If a `message` is available.
```
let text: UITextView?

// init

if let message = message {
    text = UITextView(frame: .zero)
    text.text = message
    text.font = UIFont.systemFont(ofSize: 14)
   // etc.
}
```

Sources/UIAlertControllerView.swift
```diff
@@ -47,18 +51,17 @@ class UIAlertControllerView: UIView {
         layer.cornerRadius = 3
 
         header.sizeToFit()
-        header.frame.height += verticalPadding
         header.layer.contentsGravity = .top
         addSubview(header)
+        addSubview(text)
```
mouselangelo, line 56: We add the view to the hierarchy here - but in the `layoutAsActionSheet()` case - we don't layout this view.

Maybe we don't want to implement this at the moment - in which case we could consider adding an assertion when `style == . actionSheet`.

Alternatively we could render this. I'm not sure if the native android Dialogs do actually allow this.

Sources/UIAlertControllerView.swift
```diff
@@ -69,22 +72,55 @@ class UIAlertControllerView: UIView {
...
+            let (index, button) = arg
+            button.sizeToFit()
+            button.frame.width = max(button.frame.width, 75)
+            button.contentHorizontalAlignment = .center
+            button.frame.minY = text.frame.maxY + verticalPadding
+            if button == buttons.first {
+                button.frame.origin.x = bounds.maxX - button.frame.width - verticalPadding
+            } else {
+                let previousElement = buttons[index - 1]
+                button.frame.origin.x = previousElement.frame.minX - button.frame.width
```
mouselangelo, line 100: nit / optional:

```suggestion
                button.frame.origin.x = previousElement.frame.minX - button.frame.width - verticalPadding / 2
```

The buttons are right next to each other - perhaps some gap between them is better?

Screenshot 2021-12-10 at 18 35 18

Screenshot 2021-12-10 at 18 36 04

Sources/UIAlertControllerView.swift
```diff
@@ -69,22 +72,55 @@ class UIAlertControllerView: UIView {
...
-        let elementWidth = widestElement?.frame.width ?? header.frame.width
-        let elementHeight = header.frame.height + buttons.reduce(0, { $0 + $1.bounds.height })
-
+        
+        let width: CGFloat
+        let height: CGFloat
+        
+        switch style {
+        case .actionSheet:
+            width = ([text] + buttons).map { $0.frame.width }.max()!
```
mouselangelo, line 113: nit:
Currently `text` is always size `zero` in style `actionSheet`. In case we decide not to render it - it can be excluded from this calculation (perhaps).

Sources/UIAlertControllerView.swift
```diff
@@ -69,22 +72,55 @@ class UIAlertControllerView: UIView {
...
+        case .alert:
+            let buttonsWidth = buttons.reduce(0, { $0 + $1.frame.width })
+            width = ([header.frame.width, text.frame.width, buttonsWidth]).max()!
+            height = buttons.first?.frame.maxY ?? text.frame.height
+        }
+        
         return CGSize(
-            width: max(minWidth, elementWidth) + horizontalPadding, // on right
-            height: elementHeight + verticalPadding * 2 // at bottom
+            width: max(minWidth, width) + horizontalPadding,
```
mouselangelo, line 122: Should we also consider a `maxWidth` - I guess at max the device width (minus some padding).
Right now the view could in theory extend beyond the width of the screen.

Also regardless of the width of the alert the `TextView` width remains unchanged:

Screenshot 2021-12-10 at 18 30 25

Sources/UIAlertControllerView.swift
mouselangelo: BTW - we only layout the buttons horizontally.
Hopefully for our current use case this works fine.
On Android usually when the width of the buttons get larger the buttons[ layout switches to vertical](https://material.io/components/dialogs#usage):

samples/getting-started/DemoApp/ViewController.swift
mouselangelo: I think we don't need to make these members of the `ViewController` though.

In fact they can be initialised only when needed (in the button handlers).
Ex:
```
private func displayAlertController() {
    let alertController = UIAlertController(...)
   alertController.addAction(...)
   ....
   self.present(alertController, animated: true)
}

// handling actions
buttonForAlert.onPress = {
   self.displayAlertController()
}
```

mouselangelo: I think this might also fix the issue you were facing with the alert not being displayed a second time.

mouselangelo (commented):
@rikner thanks for your work on this. It looks quite good already. I added some comments.
I think we can still improve the layout logic and make it more robust (for longer button labels, titles, texts, etc.).
Happy to pair-program with you on this next week if you like :)

rikner:
@mouselangelo Did some more cleanup and minor improvements since our pair coding session (everything from 767ee3d and later). I also tried out the styles from [this Figma page](https://www.figma.com/file/0uQd8yk8XFVplYbrNP27iw/%26-%2F-Alerts-%26-Dialogs-(Community)?node-id=0%3A1), but reverted them again since they result in a different appearance than in Figma and I thought it's not really part of this PR anyway.

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,103 @@ class UIAlertControllerView: UIView {
...
+        buttons.forEach { addSubview($0) }
+
+        subviewWidth = {
+            let minWidth: CGFloat = 300
+            let maxWidth: CGFloat
+            if UIScreen.main.isPortrait {
+                maxWidth = UIScreen.main.bounds.width - 4 * horizontalPadding
+            } else {
+                maxWidth = max(minWidth, UIScreen.main.bounds.width / 1.5)
+            }
```
mouselangelo, line 75: I think in both these cases (but especially landscape) we should also apply a max readable width. For a font-size of 16px this is approximately 540px.

```suggestion
            let minWidth: CGFloat = 300
            let maxReadableWidth: CGFloat = 540
            let maxWidth: CGFloat
            if UIScreen.main.isPortrait {
                maxWidth = min(UIScreen.main.bounds.width - 4 * horizontalPadding, maxReadableWidth)
            } else {
                maxWidth = min(max(minWidth, UIScreen.main.bounds.width / 1.5), maxReadableWidth)
            }
```

The max readable width is important as when the text gets wider than that - it's very hard to read. I'll be presenting this in my Typography dev talk - hopefully in January.

Without max readable width
too-wide

After applying the max readable width
just-right

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,103 @@ class UIAlertControllerView: UIView {
...
+            return min(maxWidth, max(largestSubviewWidth, minWidth))
+        }()
     }
 
     override func layoutSubviews() {
-        switch style {
-        case .actionSheet: layoutAsActionSheet()
-        default: assertionFailure(
-            "The UIAlertControllerStyle style, \(style), is not implemented yet!")
+        subviews.forEach { $0.sizeToFit() }
```
mouselangelo, line 87: Not 100% sure - but I think this might not be needed.

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,103 @@ class UIAlertControllerView: UIView {
...
+        return layoutButtonsVertically()
+
     }
 
-    private func layoutAsActionSheet() {
-        header.frame.origin = CGPoint(x: horizontalPadding, y: verticalPadding)
+    private func layoutButtonsVertically() {
+        buttons.enumerated().forEach({ (arg) in
+            let (index, button) = arg
+
```
mouselangelo, line 112: nit:

```suggestion
        buttons.enumerated().forEach({ (index, button) in
```

Also in the function below.

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,103 @@ class UIAlertControllerView: UIView {
...
+            case .actionSheet:
+                button.titleLabelOriginX = horizontalPadding
+            case .alert:
+                button.titleLabelOriginX = fullButtonWidth - horizontalPadding - button.bounds.width
+            }
+
+            button.frame.size.width = fullButtonWidth
+            button.frame.size.height += verticalPadding
+
+            if button == buttons.first {
```
mouselangelo, line 125: nit:
We also have the `index` so this could also be:
```suggestion
            if index == 0 {
```

However, whichever you prefer. I find this equally easy to read. Also in the function below.

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,103 @@ class UIAlertControllerView: UIView {
...
+    private func layoutButtonsHorizontally() {
         buttons.enumerated().forEach({ (arg) in
             let (index, button) = arg
-            let previousElement = (index > 0) ? buttons[index - 1] : header
             button.sizeToFit()
-            button.contentHorizontalAlignment = .left
-            button.bounds.size.height += verticalPadding
-            button.frame.size.width = max(bounds.size.width, button.frame.size.width)
-            button.frame.origin = CGPoint(x: 0, y: previousElement.frame.maxY)
+            button.clipsToBounds = true
```
mouselangelo, line 139: Just wondering if we should add the `button.clipsToBounds = true` for all buttons in vertical layout as well.
Also if it makes sense to add it all the subviews

Sources/UIAlertControllerView.swift
```diff
@@ -97,9 +166,18 @@ class UIAlertControllerButton: Button {
...
 
+    var titleLabelOriginX: CGFloat?
+    
     override func layoutSubviews() {
         super.layoutSubviews()
+        if let titleLabelOriginX = titleLabelOriginX {
+            titleLabel?.frame.origin.x = titleLabelOriginX
+        }
+    }
+}
```
mouselangelo, line 177: I guess instead of making it an optional - could make it `0` by default.

```suggestion
    var titleLabelOriginX: CGFloat = 0
    
    override func layoutSubviews() {
        super.layoutSubviews()
        titleLabel?.frame.origin.x = titleLabelOriginX
    }
}
```

samples/getting-started/DemoApp/ViewController.swift
```diff
@@ -19,8 +26,63 @@ class ViewController: UIViewController {
...
+
+        buttonForActions.setTitle("Show Actions", for: .normal)
+        buttonForActions.titleLabel?.font = .systemFont(ofSize: 20)
+        buttonForActions.sizeToFit()
+        buttonForActions.center.x = view.frame.midX
+        buttonForActions.center.y = buttonForAlert.frame.maxY + 10
+        
+        #if os(iOS)
+        buttonForAlert.addTarget(self, action: #selector(objc_presentAlertController), for: .touchDown)
+        buttonForActions.addTarget(self, action: #selector(objc_presentActionsController), for: .touchDown)
```
mouselangelo, line 44: nit:
Normally we add the touch handlers on iOS for `.touchUpInside` - this allows the user to visually see that the button is pressed (as the color changes)

mouselangelo (approved):
@rikner - this looks very good. The code is much cleaner now too. Thanks for reworking on this. Very pleased with the result.
I added a few comments - but nothing blocking IMO. 💪

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,103 @@ class UIAlertControllerView: UIView {
...
+                let previousElement = buttons[index - 1]
+                button.frame.origin.y = previousElement.frame.maxY
+            }
+
+        })
+    }
+
+    private func layoutButtonsHorizontally() {
         buttons.enumerated().forEach({ (arg) in
             let (index, button) = arg
```
rikner, line 138:
```suggestion
        buttons.enumerated().forEach({ (index, button) in
```

Sources/UIAlertControllerView.swift
rikner: I think this won't work because setting `Button.contentHorizontalAlignment` wouldn't have any effect anymore because the label will always start at 0.

Sources/UIAlertControllerView.swift
rikner: hmmm yeah this is kind of tricky. it shouldn't be necessary but it currently is for the layout to look correct. will investigate this.

rikner: (done, wasn't that complicated in the end)

Sources/UIAlertControllerView.swift
```diff
@@ -24,6 +27,17 @@ class UIAlertControllerView: UIView {
         self.style = style
         header.text = title
         header.font = UIFont.boldSystemFont(ofSize: 20)
+        header.numberOfLines = 0
```
ephemer, line 30: I guess I'm wrong but I thought 0 was the default?

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,100 @@ class UIAlertControllerView: UIView {
...
-        header.sizeToFit()
-        header.frame.height += verticalPadding
-        header.layer.contentsGravity = .top
         addSubview(header)
+        text.map { addSubview($0) }
+        buttons.forEach { addSubview($0) }
+
+        subviewWidth = {
+            let minWidth: CGFloat = 300
+            let maxReadableWidth: CGFloat = 540
```
ephemer, line 70: What is this based on? Is it related to the default font size vs the line width? I would love to see this as a product of some other known quantities rather than something hard coded. I don't consider this to be a blocker but it would be great if we could get more insight in the code as to why this value is what it is

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,100 @@ class UIAlertControllerView: UIView {
...
 
-        header.sizeToFit()
-        header.frame.height += verticalPadding
-        header.layer.contentsGravity = .top
         addSubview(header)
+        text.map { addSubview($0) }
+        buttons.forEach { addSubview($0) }
+
+        subviewWidth = {
+            let minWidth: CGFloat = 300
```
ephemer, line 69: i'm less concerned about minWidth but optionally it would be awesome to calculate this based on some other known quantity as well

Sources/UIAlertControllerView.swift
```diff
@@ -43,48 +57,100 @@ class UIAlertControllerView: UIView {
...
 
     override func layoutSubviews() {
-        switch style {
-        case .actionSheet: layoutAsActionSheet()
-        default: assertionFailure(
-            "The UIAlertControllerStyle style, \(style), is not implemented yet!")
+        header.frame.origin = CGPoint(x: horizontalPadding, y: verticalPadding)
+        header.bounds.width = subviewWidth
+        header.sizeToFit()
+        header.textAlignment = .left // sizeToFit() seems to change textAlignment
```
ephemer, line 93: can you elaborate on this comment? This sounds like a bug in our code. Or does it happen on iOS as well?

Sources/UIAlertControllerView.swift
```diff
@@ -97,9 +163,18 @@ class UIAlertControllerButton: Button {
         }
     }
 
+    var titleLabelOriginX: CGFloat?
```
ephemer, line 166: can you add a code comment describing why we need this?

ephemer (approved):

Sources/UIAlertControllerView.swift
rikner: UILabel has indeed a default value of `1`. https://github.com/flowkey/UIKit-cross-platform/blob/31b422b2553b9ea2eef1d8d398ae4a80aebde9c2/Sources/UILabel.swift#L24

Sources/UIAlertControllerView.swift
rikner: @mouselangelo suggested this value to have something to limit the width of the text. I think ideally we would come up with a value that's based on the recommended line length of 50-75 characters ([src](https://baymard.com/blog/line-length-readability))

Sources/UIAlertControllerView.swift
rikner: I checked this again. It seems it was an unnecessary comment, since 
```
header.textAlignment = .left
header.sizeToFit()
```
seems to work as expected. changing it.

Sources/UIAlertControllerView.swift
ephemer: Thanks for adding the comment! Currently it seems to be a "what" comment rather than "why" though – I still don't understand why this is necessary. Could you write the comment in terms of "what would happen if we didn't have this"?

Sources/UIAlertControllerView.swift
ephemer: thanks for clarifying and sorry for the noise!

Sources/UIAlertControllerView.swift
ephemer: Ok understood! Sounds like this would vary based on the font size of the labels then? AFAIK devs can adjust the font size of these labels so it would be great if you could at least add a comment stating that this value assumes the default font size of 12 (?) We don't have to implement this "perfectly" but it would be valuable to clarify our assumptions in the code

mouselangelo: Yes - this is very much tied to the font-size and the line-height and actually also the font itself.

PR #353, opened 2021-12-22
Use setZOrderMediaOverlay instead of setZOrderOnTop

rikner:
**Type of change:** bug fix

Allow views from react-native to be rendered on top of UIKit Views.

### DIFF
--- a/src/main/java/org/libsdl/app/SDLActivity.kt
+++ b/src/main/java/org/libsdl/app/SDLActivity.kt
@@ -82,7 +82,7 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
 
         // Set up the surface
         mSurface = SurfaceView(context)
-        mSurface.setZOrderOnTop(true) // so we can cover the video (fixes Android 8 bug)
+        mSurface.setZOrderMediaOverlay(true) // so we can cover the video (fixes Android 8 bug)
 
         // Enables the alpha value for colors on the SDLSurface
         // which makes the VideoJNI SurfaceView behind it visible


src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -82,7 +82,7 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
 
         // Set up the surface
         mSurface = SurfaceView(context)
-        mSurface.setZOrderOnTop(true) // so we can cover the video (fixes Android 8 bug)
+        mSurface.setZOrderMediaOverlay(true) // so we can cover the video (fixes Android 8 bug)
```
ephemer, line 85: I'd have to look in the git history to double check but I'm pretty sure `mSurface.setZOrderMediaOverlay(true)` is what we had _before_ we implemented this change.

We still support Android 8, so I don't think we can just change this code and expect it to work (and in particular it's dangerous to leave the comment unchanged). I would love to drop Android 8 support but that needs to be a separate discussion.

Unfortunately, changing this is a non-starter while we still support Android 8. We need to find another solution here

ephemer (changes requested):

src/main/java/org/libsdl/app/SDLActivity.kt
rikner: [This](https://github.com/flowkey/UIKit-cross-platform/pull/158/files) is the PR that introduced `setZOrderOnTop`, before that we didn't had this at all. But since I don't have an Android 8 device right now I can't say for sure that `setZOrderMediaOverlay` works on Android 8, but it does on Android 9 and Android 10 so far

ephemer: Thanks for finding this @rikner, that's really interesting! I don't know if we used ZOrderMediaOverlay in a branch or in a version that we never published to git but I'm pretty sure we have used it before and had other issues with it (maybe the decision not to use it was simply based on the docs, I don't remember 100%).

In any case, this affects rendering of the entire view hierarchy – maybe we can investigate it after the currently release but I'd rather not make the current release dependent on it :v:

michaelknoch: I'm pretty sure we haven't used `setZOrderMediaOverlay` before. We spoke about it a few times but AFAIK we never used it. I can easily reproduce the buggy behaviour with the video overlaying the sheet even on an android 12 pixel 5 emulator after removing `mSurface.setZOrderOnTop(true)`. `setZOrderMediaOverlay(true)` seems to reliably fix the bug and allows drawing things above the UIKitView (from react native for example). I agree that we should not make the release depend on it right now, but it could be a viable option to solve the issue @rikner @ephemer .

Screenshot 2021-12-22 at 19 12 23
Screenshot 2021-12-22 at 19 02 05

rikner: I made some screenshots on an Android 8 device with the various settings [in our notion docs](https://www.notion.so/flowkey/SDLActivity-Surface-Z-Order-0ca16bb0d184424795237112c8744b9e). From these I would say that we can use `setZOrderMediaOverlay` unless anybody is aware of cases where something else would break (e.g. we talked about some background colour, but not sure yet what exactly is meant).

michaelknoch:
closing this because it was part of https://github.com/flowkey/UIKit-cross-platform/commit/6c78d1dada9b8246500a1b00ffa3162f251edb84 already

PR #354, opened 2022-02-02
Downgrade ExoPlayer to v2.13.2 to match react-native-video version

mouselangelo:
See: https://github.com/flowkey/mobile-app/pull/2254
See https://github.com/flowkey/NativePlayer/pull/939

**Type of change:** chore

The latest `react-native-video` uses ExoPlayer `v2.13.2` - while our current `fix-touch` branch uses `v2.13.3`- it might be good to align these versions.

Note that our current production app uses `v2.11.7` so we are not really rolling back to an older version as far as users are concerned.

[Release notes for v2.13.3](https://github.com/google/ExoPlayer/releases/tag/r2.13.3)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "com.google.android.exoplayer:exoplayer:2.13.3"
+    implementation "com.google.android.exoplayer:exoplayer:2.13.2"
 }
 
 repositories {


rikner (approved):

michaelknoch:
Interesting, so it was even more out of sync but it didn't crash

ephemer (approved):

mouselangelo:
> Interesting, so it was even more out of sync but it didn't crash

@michaelknoch - so the reason it was crashing is that v2.13.2 had some breaking changes in the API from the previous v2.11.7.  You can see this in https://github.com/flowkey/UIKit-cross-platform/commit/f8cb41875a3d7a6fa726170e48c68cf0498bd5ac

They introduced the `Builder` pattern and deprecated some constructors and also changed some methods. Surprisingly they just removed some methods that could be previously used.
Example:
The call to `AdaptiveTrackSelection.Factory(bandwidthMeter)` - no longer accepts `bandwidthMeter`

Erik had already fixed this - but we are just using an older commit in the NativePlayer- which should be resolved by https://github.com/flowkey/NativePlayer/pull/939

PR #357, opened 2022-03-10
Add onVideoBuffering callback

rikner:
Fixes # 

**Type of change:** 

We want to provide a way to know whether the video is buffering.

### DIFF
--- a/Sources/AVPlayer+Android.swift
+++ b/Sources/AVPlayer+Android.swift
@@ -10,7 +10,9 @@ import JNI
 
 public class AVPlayer: JNIObject {
     public var onLoaded: ((Error?) -> Void)?
+    public var onVideoReady: (() -> Void)?
     public var onVideoEnded: (() -> Void)?
+    public var onVideoStalled: (() -> Void)?
 
     public convenience init(playerItem: AVPlayerItem) {
         let parentView = JavaSDLView(getSDLView())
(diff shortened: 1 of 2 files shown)


Sources/AVPlayer+Android.swift
```diff
@@ -10,7 +10,9 @@ import JNI
 
 public class AVPlayer: JNIObject {
     public var onLoaded: ((Error?) -> Void)?
+    public var onVideoReady: (() -> Void)?
     public var onVideoEnded: (() -> Void)?
+    public var onVideoStalled: (() -> Void)?
```
mouselangelo, line 15: nit:
I'd prefer the callback to be named similar to the actual event so `onVideoBuffering`

mouselangelo (approved):
@rikner - overall change looks great. I added a comment about the name of the callback(s), but consider it non-blocking

Sources/AVPlayer+Android.swift
rikner: Ah right, the reason why I named it `onVideoStalled` was to resemble [AVPlayerItemPlaybackStalled](https://developer.apple.com/documentation/foundation/nsnotification/name/1387661-avplayeritemplaybackstalled). But now that I look at the code again there is no reason to do that. Will change it 👍

PR #360, opened 2022-03-18
complete pending animations before UIScreen gets deinited

michaelknoch:
**Type of change:** bugfix

Complete pending animations before UIScreen gets deinited. This corresponds to iOS behaviour.

**Expectation:**
- start animation with any duration/delay
- move app in background and foreground again
- animations are completed

## iOS behaviour

```swift
let box = UIView()
box.backgroundColor = .red
box.frame = CGRect(x: 0, y: 100, width: 200, height: 200)

UIView.animate(withDuration: 15, delay: 5, options: [], animations: {
    box.frame = CGRect(x: 300, y: 400, width: 200, height: 300)
    box.backgroundColor = .blue
}, completion: { finished in
    print("finished", finished)
})
```

https://user-images.githubusercontent.com/5617793/159062807-ed27cb31-e12a-47ea-b836-ad9551d37b01.mp4

even skips the delay:

https://user-images.githubusercontent.com/5617793/159062792-d96b89fa-3607-4379-8e14-2284148d8e09.mp4

### DIFF
--- a/Sources/UIScreen.swift
+++ b/Sources/UIScreen.swift
@@ -106,7 +106,6 @@ public final class UIScreen {
         UIView.completePendingAnimations()
         UIView.layersWithAnimations.removeAll()
         UIView.currentAnimationPrototype = nil
-
         UIEvent.activeEvents.removeAll()
         FontRenderer.cleanupSession()
 
--- a/Sources/UIScreen.swift
+++ b/Sources/UIScreen.swift
@@ -103,9 +103,11 @@ public final class UIScreen {
     }
 
     deinit {
+        UIView.completePendingAnimations()
         UIView.layersWithAnimations.removeAll()
-        UIEvent.activeEvents.removeAll()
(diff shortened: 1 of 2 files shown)


rikner (approved):
I don't understand what exactly we're achieving with `animate(at: Timer(startingAt: NSDate.distantFuture.timeIntervalSince1970)` but I see that it fixes the issue with animations getting "stuck" 💯

michaelknoch:
ok fair point, I will add a comment. Before we remove all the pending animations in `deInit` of `UIScreen` we run them with a date in the future to ensure the animated layers represent the state of the finished animation. I saw the same behaviour on iOS. @rikner

ephemer (approved):
LGTM! The only thing I'm not too stoked about is adding another Foundation import, but that is a problem for another day. Thanks a lot for investigating and fixing this @michaelknoch !

PR #362, opened 2022-04-02
make sure Timer does not crash for huge startingTimeInMilliseconds values on 32 bit arch

michaelknoch:
**Type of change:** bugfix

- fixes [this crash](https://app.bugsnag.com/flowkey-gmbh/mobile-app/errors/623efdd47648f400086e64de?filters[app.release_stage]=production&filters[release.seen_in][]=2.35.4&filters[release.seen_in][]=2.35.4%20(*)&filters[event.unhandled]=true)
- reverts the first attempt to fix the crash https://github.com/flowkey/UIKit-cross-platform/commit/6cb855df84e9b98638321cdfe87b547b04fc4e91
- improves timer to ensure it does not crash for huge `startingTimeInMilliseconds` values on 32 bit arch

### DIFF
--- a/Sources/Timer.swift
+++ b/Sources/Timer.swift
@@ -20,12 +20,8 @@ struct Timer {
 
         let seconds = (startingTimeInMilliseconds / 1000).rounded(.down)
         let milliseconds = startingTimeInMilliseconds.truncatingRemainder(dividingBy: 1000)
-        if !seconds.isZero {
-            startTime.tv_sec += type(of: startTime.tv_sec).init(seconds)
-        }
-        if !milliseconds.isZero {
-            startTime.tv_usec += type(of: startTime.tv_usec).init(milliseconds * 1000)
-        }
+        startTime.tv_sec += type(of: startTime.tv_sec).init(seconds)
+        startTime.tv_usec += type(of: startTime.tv_usec).init(milliseconds * 1000)
 
         self.startTime = startTime
     }
--- a/Sources/Timer.swift
+++ b/Sources/Timer.swift
@@ -18,13 +18,13 @@ struct Timer {
         var startTime = timeval()
         gettimeofday(&startTime, nil)
 
-        let seconds = startingTimeInMilliseconds / 1000
+        let seconds = (startingTimeInMilliseconds / 1000).rounded(.down)
         let milliseconds = startingTimeInMilliseconds.truncatingRemainder(dividingBy: 1000)
         if !seconds.isZero {
             startTime.tv_sec += type(of: startTime.tv_sec).init(seconds)
         }
         if !milliseconds.isZero {
-            startTime.tv_usec += type(of: startTime.tv_usec).init(milliseconds)
+            startTime.tv_usec += type(of: startTime.tv_usec).init(milliseconds * 1000)
         }
 
         self.startTime = startTime
--- a/Sources/Timer.swift
+++ b/Sources/Timer.swift
@@ -17,9 +17,16 @@ struct Timer {
     init(startingAt startingTimeInMilliseconds: Double = 0.0) {
         var startTime = timeval()
         gettimeofday(&startTime, nil)
-        if !startingTimeInMilliseconds.isZero {
-            startTime.tv_usec += type(of: startTime.tv_usec).init(startingTimeInMilliseconds * 1000)
+
+        let seconds = startingTimeInMilliseconds / 1000
+        let milliseconds = startingTimeInMilliseconds.truncatingRemainder(dividingBy: 1000)
+        if !seconds.isZero {
+            startTime.tv_sec += type(of: startTime.tv_sec).init(seconds)
         }
+        if !milliseconds.isZero {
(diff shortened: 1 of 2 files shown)


Sources/UIView+animate.swift
```diff
@@ -73,9 +73,7 @@ extension UIView {
 
     static func completePendingAnimations() {
         layersWithAnimations.forEach {
-            // run pending animations which are supposed to finish within the next 60 minutes
-            let oneHourInMs: Double = 60 * 60 * 1000
```
michaelknoch, line 77: it turned out `60 * 60 * 1000` is still bigger than a long which is a Int32 on 32bit arch. @ephemer please correct me if this is wrong

Sources/Timer.swift
```diff
@@ -17,9 +17,16 @@ struct Timer {
     init(startingAt startingTimeInMilliseconds: Double = 0.0) {
         var startTime = timeval()
         gettimeofday(&startTime, nil)
-        if !startingTimeInMilliseconds.isZero {
-            startTime.tv_usec += type(of: startTime.tv_usec).init(startingTimeInMilliseconds * 1000)
```
michaelknoch, line 21: `tv_usec` overflows on arch32 when passing in a value which is bigger than `2147483647` which means that the limit of startingTime is ~34 minutes

Sources/UIView+animate.swift
michaelknoch: we changed `NSDate.distantFuture.timeIntervalSince1970` to `60 * 60 * 1000` which is still too big

Sources/Timer.swift
```diff
@@ -17,9 +17,16 @@ struct Timer {
...
+
+        let seconds = (startingTimeInMilliseconds / 1000).rounded(.down)
```
michaelknoch, line 21: not sure about the rounding. Is there a cleaner way to split ms into s + ms?

Sources/UIView+animate.swift
ephemer: Int32.MAX == 2^31 - 1 == 2_147_483_647, which is quite a bit bigger than 3.6 million (60 * 60 * 1000). So if we're counting on that, I don't think the change in this PR will fix the issue :(

Sources/Timer.swift
```diff
@@ -17,9 +17,16 @@ struct Timer {
...
+        let milliseconds = startingTimeInMilliseconds.truncatingRemainder(dividingBy: 1000)
+        if !seconds.isZero {
```
ephemer, line 23: As an aside: it doesn't seem necessary to check whether `seconds.isZero`? We would just `0 += 0` in that case – juckt?

ephemer (approved):
I don't want to block this PR but if the issue is the 60 * 60 * 1000 thing then I'm not sure this will fix it (although it does make sense to split the time value into seconds and microseconds like we are here)

Sources/UIView+animate.swift
michaelknoch: we transform  `60 * 60 * 1000` in ms to ns by multiplying it again with 1000. this is causing the crash

60 * 60 * 1000 * 1000 which is 3_600_000_000

ephemer: ahh, ok, but that will be fixed by using whole seconds in the first instance. So the fix will still work, but the github comment is a bit confusing

michaelknoch: fair point @ephemer . I missed the conversion from ms to ns in the first comment.

michaelknoch: @ephemer do you think we should rename the timer argument `startingAt` to `startingAtFromNow`? As far as I understand `gettimeofday` sets the timer to now and then we may increment it by passing an argument. This why I changed the call to `$0.animate(at: Timer(startingAt: NSDate.distantFuture.timeIntervalSinceNow))` (instead of timeIntervalSince1970)

ephemer: I don't really have an opinion about the naming. Can you give an idea what the Timer is used for there? The API looks unfamiliar to me

rikner (commented):
@michaelknoch I could also reproduce the crash on the 32bit galaxy tab we have in the office and this fixed it for me as well 👍

Sources/UIView+animate.swift
michaelknoch: but we can discuss this in another iteration. I'm going to merge this pr now

ephemer: Ok fair enough. To be honest `animate(at:` sounds like it should be receiving a timestamp since 1970 (or on Apple platforms, since 2001) and it is pretty confusing to me that we have a `Timer` in there at all, which to me is something that should be measuring a time _interval_ (duration) rather than defining a specific point (instant) in time.

All of this will go away with the work by Swift itself on defining these concepts in the language – I would suggest we just leave it for now and use the official APIs once they become available. https://github.com/apple/swift-evolution/blob/main/proposals/0329-clock-instant-duration.md

PR #366, opened 2022-06-02
fix crash in font rendering

michaelknoch:
**Type of change:** bugfix

Fixes a crash which happens when initing UIFont while main screen doesn't exist.
See https://github.com/flowkey/NativePlayer/pull/981

### DIFF
--- a/Sources/UIFont.swift
+++ b/Sources/UIFont.swift
@@ -33,6 +33,10 @@ open class UIFont {
     }
 
     public init?(name: String, size: CGFloat) {
+        guard UIScreen.main != nil else {
+            return nil
+        }
+
         let name = name.lowercased()
         let size = Int32(size * UIScreen.main.scale)
 
--- a/Sources/UILabel.swift
+++ b/Sources/UILabel.swift
@@ -48,7 +48,8 @@ open class UILabel: UIView {
         layer.contentsGravity = textAlignment.contentsGravity()
     }
 
-    open var font: UIFont = .systemFont(ofSize: 16) {
+    static var defaultFont = UIFont.systemFont(ofSize: 16)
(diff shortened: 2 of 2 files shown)


Sources/UILabel.swift
```diff
@@ -48,7 +48,8 @@ open class UILabel: UIView {
         layer.contentsGravity = textAlignment.contentsGravity()
     }
 
-    open var font: UIFont = .systemFont(ofSize: 16) {
+    static var defaultFont = UIFont.systemFont(ofSize: 16)
+    open var font: UIFont = UILabel.defaultFont {
```
michaelknoch, line 52: I noticed that the crash actually comes from the default here

```swift
let label = UILabel(frame: frame)
label.font = NoteName.font // already static
```

ephemer:
Hi @michaelknoch let's have a quick call about this. I am skeptical that the latest PRs are the correct way to tackle the crashes we are seeing

Sources/UILabel.swift
rikner: So is this intended to do anything? I don't see a functional change. And shouldn't `defaultFont` be a `let`?

Sources/UIFont.swift
```diff
@@ -33,6 +33,10 @@ open class UIFont {
     }
 
     public init?(name: String, size: CGFloat) {
+        guard UIScreen.main != nil else {
+            return nil
+        }
```
rikner, line 38: nit: we could also do `guard let mainScreen = UIScreen.main ...` and use it below.

rikner (approved):

michaelknoch:
closing this because of https://github.com/flowkey/UIKit-cross-platform/pull/367
fyi @rikner

PR #368, opened 2022-06-10
simplify onNativeTouch method

michaelknoch:
**Type of change:** bugfix

Get rid of unused function parameters to maybe fix a crash in `onNativeTouch` function.

**trace from bugsnag** 
```
SIGTRAP Trace/breakpoint trap 
    build/arm64-v8a/<compiler-generated>:265 UIKit.onNativeTouch(env
```

**trace from google play**
```
#00  pc 00000000000a17b0  /data/app/com.flowkey.app-Cfv7n2a_ZlqbZd_l5StZdg==/lib/arm64/libUIKit.so ($s5UIKit13onNativeTouch3env4view13touchDeviceID013pointerFingerI06action1x1y8pressure11timestampMsySpySPySo18JNINativeInterfaceVGSgGSg_SvSgs5Int32VA2TS3fs5Int64VtF+244)
  #00  pc 0000000000162ee4  /data/app/com.flowkey.app-Cfv7n2a_ZlqbZd_l5StZdg==/oat/arm64/base.odex (art_jni_trampoline+180)
```

https://app.bugsnag.com/flowkey-gmbh/mobile-app/errors/61a30ddf3b2d9500071b9eea?filters[app.release_stage]=production&filters[release.seen_in][]=2.37.3&filters[release.seen_in][]=2.37.3%20(*)&filters[event.unhandled]=true

### DIFF
--- a/Sources/UIApplication+handleSDLEvents.swift
+++ b/Sources/UIApplication+handleSDLEvents.swift
@@ -252,7 +252,7 @@ public func onNativeTouch(
         SDL_TouchFingerEvent(
             type: eventType.rawValue,
             timestamp: UInt32(timestampMs),
-            touchId: Int64(touchDeviceId),
+            touchId: Int64(touchDeviceId), // some arbitrary number, stays the same per device
             fingerId: Int64(pointerFingerId),
             x: x / Float(UIScreen.lastKnownScale),
             y: y / Float(UIScreen.lastKnownScale),
--- a/Sources/UIApplication+handleSDLEvents.swift
+++ b/Sources/UIApplication+handleSDLEvents.swift
@@ -7,6 +7,7 @@
 //
 
 import SDL
+import JNI
 import struct Foundation.TimeInterval
 
 extension UIApplication {
@@ -228,16 +229,18 @@ import JNI
 public func onNativeTouch(
     env: UnsafeMutablePointer<JNIEnv?>?,
     view: JavaObject?,
-    action: JavaInt,
-    x: JavaFloat,
-    y: JavaFloat,
-    timestampMs: JavaLong
+    touchParameters: JavaObject
 ) {
     if env == nil || view == nil {
         // We're in an invalid state where the JNI has exploded somehow
         // Passing anything on to UIKit now would probably lead to a crash
         return
     }
+
+    let action: JavaInt = try! jni.GetField("action", from: touchParameters)
+    let x: JavaFloat = try! jni.GetField("x", from: touchParameters)
+    let y: JavaFloat = try! jni.GetField("y", from: touchParameters)
+    let timestampMs: JavaLong = try! jni.GetField("timestamp", from: touchParameters)
 
     guard let eventType = SDL_EventType.eventFrom(androidAction: action)
     else { return }
--- a/Sources/UIApplication+handleSDLEvents.swift
+++ b/Sources/UIApplication+handleSDLEvents.swift
@@ -228,12 +228,9 @@ import JNI
 public func onNativeTouch(
     env: UnsafeMutablePointer<JNIEnv?>?,
     view: JavaObject?,
-    touchDeviceID: JavaInt,
-    pointerFingerID: JavaInt,
     action: JavaInt,
     x: JavaFloat,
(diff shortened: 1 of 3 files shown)


ephemer:
Hey @michaelknoch I really like the idea. From memory there is potentially an arbitrary change in calling convention when a function has more than 5 arguments. With the changes here the function still takes more than 5 arguments, so I'm not sure it will make much difference as written.

Can we do this slightly differently and just pass a single object as an argument, i.e. `this.onNativeTouchUIKit(parameters)`, where `parameters` is an object containing all the existing properties (reverting the change to remove some from this PR), and unwrap them via JNI's GetField function in Swift?

Alternatively we could do this a bit lower level and use a java array (`[x, y, pressure, timestamp, etc]`) and then access / decode the array contents in Swift via JNI GetArrayContents (forget the exact name). That would be slightly more performant but also more error prone – hopefully it isn't necessary.

michaelknoch:
Makes sense @ephemer ! Seems to work with a class a0c83834be9dbc04da8a91983acbc6258206ea33. I deliberately still omitted the unused parameters because maybe the crash is just coming from one of these type castings when converting the different java types to swift types. If you are certain it can't be the cause of the crash I'm happy to add them. Otherwise I'd suggest to add them in a second iteration.

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -228,34 +229,33 @@ import JNI
...
-            touchId: Int64(touchDeviceID), // some arbitrary number, stays the same per device
-            fingerId: Int64(pointerFingerID),
+            touchId: 0,
+            fingerId: 0,
             x: x / Float(UIScreen.lastKnownScale),
             y: y / Float(UIScreen.lastKnownScale),
             dx: 0,
             dy: 0,
-            pressure: pressure
+            pressure: 0
```
michaelknoch, line 258: still omitted the unused props. See comment https://github.com/flowkey/UIKit-cross-platform/pull/368#issuecomment-1152423925

ephemer:
Hey @michaelknoch, looks great! Thanks for the update. I am fairly sure the conversion is not causing the issue (since the crash appears to occur when _calling_ the function I can't really imagine how that would happen), and if we omit the parameters now we won't know whether moving to the object structure actually fixed the issue or not. I would try a reduction in parameters as a further step if we still see the crash after this change. If it is an issue in conversions somewhere then, after this change, we'd see a crash in the `GetField` function rather than when calling the function, which would be really useful to know.

The only thing I'm wondering is how this goes performance-wise – I imagine it's fine, but I am also wondering whether we can cache some of the intermediate steps in getting the field values from this object. Were you able to test this on a device? If scrolling the sheet still feels smooth then we can just leave it as is for now

michaelknoch:
Ok fair point. I wanted to go the opposite way and try to fix it asap and then find out what actually fixed it. But this way is also fine. I agree that it's very unlikely that the crash comes from the conversions. I added the remaining fields: https://github.com/flowkey/UIKit-cross-platform/pull/368/commits/10c23817ffddac6f30b82c62965dc9318481800e

Performance seems to be fine on my device. Touches never feel snappy to me on android and it's hard to tell if it got worse with this change. I don't think so. @ephemer @rikner

ephemer:
Looks great to me @michaelknoch! Hope we can finally stamp out this crash once and for all 🙏🏼

ephemer (approved):
🙌🏼

src/main/java/org/libsdl/app/SDLOnTouchListener.kt
```diff
@@ -85,3 +86,13 @@ private fun MotionEvent.touchValues(i: Int): TouchValues {
             min(this.getPressure(i), 1.0f)
     )
 }
+
+class TouchParameters(
```
rikner, line 90: nit: not sure if it makes any difference to use a [`data class` ](https://kotlinlang.org/docs/data-classes.html) here instead. would be slightly more kotlin-idiomatic since we're really just storing some data here.

rikner (approved):
nice! really hope that does the trick 🙏

PR #373, opened 2022-07-26
fix bottom insets when using non translucent navigation

michaelknoch:
**Type of change:** UX

Only includes insets for navigation bars when navigation color is set to transparent.

### DIFF
--- a/src/main/java/org/libsdl/app/SDLActivity.kt
+++ b/src/main/java/org/libsdl/app/SDLActivity.kt
@@ -104,12 +104,13 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
         val zeroRect = RectF(0f, 0f, 0f, 0f)
 
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-            val insets = rootView.rootWindowInsets?.getInsets(
-                WindowInsets.Type.displayCutout() or
-                WindowInsets.Type.navigationBars()
-            ) ?: return zeroRect
+            val activity = context as Activity
+            val typeMask =
+                if (activity.window.navigationBarColor == Color.TRANSPARENT) WindowInsets.Type.displayCutout() or WindowInsets.Type.navigationBars()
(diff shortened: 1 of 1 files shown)


src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -104,12 +104,13 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
         val zeroRect = RectF(0f, 0f, 0f, 0f)
 
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-            val insets = rootView.rootWindowInsets?.getInsets(
-                WindowInsets.Type.displayCutout() or
-                WindowInsets.Type.navigationBars()
-            ) ?: return zeroRect
+            val activity = context as Activity
+            val typeMask =
+                if (activity.window.navigationBarColor == Color.TRANSPARENT) WindowInsets.Type.displayCutout() or WindowInsets.Type.navigationBars()
```
michaelknoch, line 109: this is a hack, it's not only the color which has to be set to make content appear behind the system navigation, but I found this to be the most reliable way to.

michaelknoch: In our case we ask for navigation insets for devices with gesture navigation enabled

ephemer (approved):

rikner:
What exactly is the underlying problem?
Is it that some android devices don't make the content appear behind the navigation bar although we tell them too?

michaelknoch:
The problem is that the `getInsets()` method returns a bottom inset for the navigation (~40px) although the content does **not** draw behind it. In that case we shouldn't have to handle the bottom safe area on our own, the system renders the app above the navigator and should return a bottom inset of 0. I think this is a bug with the getInsets method to be honest @rikner

rikner (approved):

michaelknoch:
thanks for the quick review guys @rikner @ephemer 🥳 !

PR #376, opened 2022-11-04
Remove Foundation

ephemer:
**Type of change:** DX, Perf, Bugfix

Foundation is large and buggy on non-Darwin platforms. This PR follows on from #344 to continue removing the last dependencies we have on the library (and also on Dispatch wherever possible).


Sources/CGFloat.swift
```diff
@@ -1,10 +1 @@
...
-//  CGFloat.swift
-//  UIKit
-//
-//  Created by Chris on 19.06.17.
-//  Copyright © 2017 flowkey. All rights reserved.
-//
-
-import struct Foundation.CGFloat
-public typealias CGFloat = Foundation.CGFloat
+public typealias CGFloat = Double
```
ephemer, line 1: since https://github.com/apple/swift-evolution/blob/main/proposals/0307-allow-interchangeable-use-of-double-cgfloat-types.md, trying to mimic the original behaviour does more harm than good. It is widely accepted that `Double` is a better type to use by default, even on arm32.

Sources/UIApplication.swift
```diff
@@ -87,12 +86,19 @@ import JNI
 
 private let maxFrameRenderTimeInSeconds = 1.0 / 60.0
 
+// ******************
+// Requires Dispatch
+import Dispatch
```
ephemer, line 91: Without this, we need CFRunLoopRun to drain the main queue. This is a major breakthrough for the work at hand.

michaelknoch (approved):

PR #379, opened 2022-12-05
process pending events/tasks after `nativeDestroyScreen`

michaelknoch:
**Type of change:** bugfix

process pending events/tasks after `nativeDestroyScreen`

Screenshot 2022-12-05 at 01 16 58


src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -361,6 +361,9 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
         onNativeSurfaceDestroyed()
         nativeDestroyScreen()
         removeFrameCallback()
+
+        // renderer should now be destroyed but we need to process events once more to clean up
+        this.nativeProcessEventsAndRender()
```
michaelknoch, line 366: we do the same already in `removeFrameCallbackAndQuit`

PR #381, opened 2022-12-07
fix a crash in UIscreen init/deinit

michaelknoch:
**Type of change:** bugfix

We experience a crash when destroying and to recreating UIScreen a few times. It happens roughly 1 out of 10 times. It's harder to reproduce in a debug build.

```
SIGTRAP: Trace/breakpoint trap

0 libswiftCore.so +0x1fa800 $ss17_assertionFailure__4file4line5flagss5NeverOs12StaticStringV_SSAHSus6UInt32VtF
1 libUIKit.so +0x1b3fb4     UIKit.UIScreen.__allocating_init() -> UIKit.UIScreen (UIScreen.swift:74)
2 libUIKit.so +0x1732dc     UIKit.nativeInit(env (androidNativeInit.swift:33)
3 libart.so +0x440550       art_quick_generic_jni_trampoline
4 libart.so +0x20a2ac       nterp_helper
5 base.odex +0x193f53a      0x7174a1853e
6 libart.so +0x20a250       nterp_helper
7 base.odex +0x193fac2      0x7174a18ac6
8 boot.oat +0xe8909c        android.view.SurfaceView.updateSurface
```
https://app.bugsnag.com/flowkey-gmbh/mobile-app/errors/638a30164fb0c8000955fc8c?filters[app.release_stage]=production&filters[release.seen_in]=2.45.0%20(2154926)&filters[event.unhandled]=true&filters[user.id]=Nhf6LvAP8oWmYDKF6

### DIFF
--- a/Sources/AVPlayerLayer+Android.swift
+++ b/Sources/AVPlayerLayer+Android.swift
@@ -14,6 +14,7 @@ public enum AVLayerVideoGravity: JavaInt {
     case resizeAspectFill = 4 // RESIZE_MODE_ZOOM
 }
 
+@MainActor
 public class AVPlayerLayer: JNIObject {
     override public static var className: String { "org.uikit.AVPlayerLayer" }
 
@@ -31,15 +32,13 @@ public class AVPlayerLayer: JNIObject {
     public var frame: CGRect {
         get { return .zero } // FIXME: This would require returning a JavaObject with the various params
         set {
-            Task { @MainActor in
-                let scaledFrame = (newValue * UIScreen.main.scale)
-                try! call(methodName: "setFrame", arguments: [
-                    JavaInt(scaledFrame.origin.x.rounded()),
-                    JavaInt(scaledFrame.origin.y.rounded()),
-                    JavaInt(scaledFrame.size.width.rounded()),
-                    JavaInt(scaledFrame.size.height.rounded())
-                ])
-            }
+            let scaledFrame = (newValue * UIScreen.main.scale)
+            try! call(methodName: "setFrame", arguments: [
+                JavaInt(scaledFrame.origin.x.rounded()),
+                JavaInt(scaledFrame.origin.y.rounded()),
+                JavaInt(scaledFrame.size.width.rounded()),
+                JavaInt(scaledFrame.size.height.rounded())
+            ])
         }
     }
 
--- a/src/main/java/org/libsdl/app/SDLActivity.kt
+++ b/src/main/java/org/libsdl/app/SDLActivity.kt
@@ -1,17 +1,18 @@
 package org.libsdl.app
 
 import android.app.Activity
-import java.lang.reflect.Method
-
-import android.view.*
-import android.widget.RelativeLayout
-import android.util.Log
-import android.graphics.*
-import android.view.KeyEvent.*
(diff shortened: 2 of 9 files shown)


src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -374,4 +375,13 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
...
     }
+
+    private fun processPendingSDLEvents(invocations: Int = 0) {
+        if (invocations > 5) return
+
+        this.nativeProcessEventsAndRender()
+        Handler(Looper.getMainLooper()).postDelayed({
+            processPendingSDLEvents(invocations + 1)
+        }, 100)
+    }
```
michaelknoch, line 386: I don't know if this has any effect to be honest. I felt like it makes things slightly better but the crash still exists (without the changes in UIScreen.swift)

ephemer: Would rather not some arbitrary timeouts like this. This may fix some things but hide others. We want to fix these issues predictably and deterministically rather than randomly.

ephemer (approved):
The change in UIScreen makes a lot of sense to me, but the change in SDLActivity doesn't – would you be fine reverting the latter change and just merging the first one?

We could use something similar to what we do in the JNI library:

```
if isMainThread {
  cleanUpSingletons()
} else {
  DispatchQueue.main.sync(cleanUpSingletons)
}
```

Note that `isMainThread` is exposed when importing `JNI`.

We could also consider just running `_dispatch_main_queue_callback_4CF` manually in UIScreen.deinit after queuing the Task, since that's what we were trying to achieve with the changes in SDLActivity as written.

Sources/AVPlayerLayer+Android.swift
```diff
@@ -31,15 +32,13 @@ public class AVPlayerLayer: JNIObject {
...
-                    JavaInt(scaledFrame.size.height.rounded())
-                ])
-            }
+            let scaledFrame = (newValue * UIScreen.main.scale)
+            try! call(methodName: "setFrame", arguments: [
+                JavaInt(scaledFrame.origin.x.rounded()),
+                JavaInt(scaledFrame.origin.y.rounded()),
+                JavaInt(scaledFrame.size.width.rounded()),
+                JavaInt(scaledFrame.size.height.rounded())
+            ])
```
michaelknoch, line 41: this change is unrelated to the crash, but I thought it's easier to flag the AVPlayerLayer with @mainactor and get rid of the task

src/main/java/org/libsdl/app/SDLActivity.kt
michaelknoch: btw, it was not the timeout which fixed the problem. Calling `this.nativeProcessEventsAndRender()` "a few times" was the fix (twice is probably good enough most of the times). I thought it's a good idea to wait a few ms inbetween to mimic the behaviour of the choreographer. Still it concerns me that tasks in deinit do not get called. I agree that our approach with the Dispatch.sync is the correct approach here, but I find it unexpected that tasks in deinit remain unexecuted @ephemer

ephemer (approved):

PR #382, opened 2022-12-07
Remove global AVPlayer instance

rikner:
Fixes # 

**Type of change:** 

We have a global AVPlayer instance in order to be able to reference our AVPlayer inside JNI callbacks. While this works it's not a good practice and wouldn't work anymore once there are more than on AVPlayer instances. 
This PR saves the memory address of our AVPlayer inside the kotlin implementation (userContext) which we can use to later retrieve our player instance in the native callbacks.


src/main/java/org/uikit/VideoJNI.kt
```diff
@@ -79,15 +82,21 @@ class AVPlayer(parent: SDLActivity, asset: AVURLAsset) {
...
+                    nativeOnVideoError(error.type, message, userContext)
+                }
             }
         }
 
         exoPlayer.addListener(listener)
     }
 
+    fun setUserContext(ctx: Long) {
+        this.userContext = ctx
```
ephemer, line 97: Can you explain the naming of `userContext`? If the idea based on C APIs which take a `void * userCtx` parameter, then I think the naming might be too generic for what it really is.

From my current understanding, I would just call this "pointerToSwiftAVPlayerInstance" or something equally specific. `userContext` in the C meaning of the word is something very generic that could be anything at all, which is why it's usually defined as `void*`. That doesn't really fit our use case here 🙏🏼

ephemer (approved):
I would like to see the naming of what is currently called `userContext` more closely reflect what it really is, rather than using the generic naming. Other than that though, this looks good to me!

michaelknoch (approved):

michaelknoch:
thanks @rikner 🙏

rikner:
Thanks @michaelknoch for getting this in.
Nearly forgot about this.

PR #383, opened 2022-12-09
Truncate timestamps if needed, start timestamps from session start rather than system uptime

ephemer:
Fixes crash in `onNativeTouch`

**Type of change:** Bug fix

The dreaded crash in `onNativeTouch` is our longest-standing and most prevalent crash on Android, and surely the most annoying too because reinstalling the app doesn’t even help, even if it was working just a moment ago.

Here’s the thing: when the user presses the screen, we need to know when that happened so we can measure things like acceleration when dragging. Android gives us that number in milliseconds since the device was switched on.

Now the problem: due to reasons out of our control, SDL only accepts values up to 2^32 (about 4.3 billion) for the timestamp. But since it’s measured since the device started and not since the app started, it’s totally expected that that number can be reached: indeed, 49.7 days after switching on the device, the number of milliseconds will be higher than the number a `UInt32` can accept without overflow. So – with the code how it is at present – every time a touch event is registered from then on, the app will crash. And no amount of restarting or reinstalling the app will fix it. Only restarting the device!

This has probably become more and more relevant over the years as devices become more stable and are left on for longer periods. I’m actually amazed that the crash doesn’t happen more often than it does!

This PR fixes that issue in two ways:

1. By initing the UInt32 with `truncatingIfNeeded` – this will ensure the conversion will never crash. Instead, it will effectively wrap around from 2^32 back to 0. That has the potential for some huge negative time deltas, which might lead to glitches, so also:
2. By basing the timestamp on the init time of `SDLActivity`, rather than on when the system booted. This will ensure there will be no weirdness if you are running the app after 49.7 days of system uptime. Conversely, it is much less likely that any given app session will be up for 49.7 days. 

I also added a defensive check to ensure that `UIScreen.main` exists before posting any events.
And I fixed some formatting weirdness left over from a previous PR – I can recommend hiding whitespace changes while reviewing this PR.

### DIFF
--- a/Sources/UIApplication+handleSDLEvents.swift
+++ b/Sources/UIApplication+handleSDLEvents.swift
@@ -236,12 +236,6 @@ public func onNativeTouch(
     view: JavaObject?,
     touchParameters: JavaObject
 ) {
-    if env == nil || view == nil {
-        // We're in an invalid state where the JNI has exploded somehow
-        // Passing anything on to UIKit now would probably lead to a crash
-        return
-    }
-
     let touchDeviceId: JavaInt = try! jni.GetField("touchDeviceId", from: touchParameters)
     let pointerFingerId: JavaInt = try! jni.GetField("pointerFingerId", from: touchParameters)
     let action: JavaInt = try! jni.GetField("action", from: touchParameters)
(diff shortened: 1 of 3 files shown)


Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -254,23 +254,27 @@ public func onNativeTouch(
...
-            touchId: Int64(touchDeviceId), // some arbitrary number, stays the same per device
-            fingerId: Int64(pointerFingerId),
-            x: x / Float(UIScreen.main.scale),
-            y: y / Float(UIScreen.main.scale),
-            dx: 0,
-            dy: 0,
-            pressure: pressure
+            SDL_TouchFingerEvent(
+                type: eventType.rawValue,
+                timestamp: UInt32(truncatingIfNeeded: timestampMs),
```
ephemer, line 264: I'm pretty sure we established this is not the cause of the crash already, but there is no harm in this change: if `timestampMs` fits inside `UInt32`, then it will remain unchanged, and if it doesn't, things will still work fine.

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -254,23 +254,27 @@ public func onNativeTouch(
     else { return }
 
     Task { @MainActor in
+        guard let screenScale = UIScreen.main?.scale else {
+            return
+        }
```
ephemer, line 259: This seems much more likely to be the cause of the crash. I don't know when it'd be possible to receive touch events when `UIScreen` is `nil`, but it seems like a no brainer not to allow that specifically.

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -236,12 +236,6 @@ public func onNativeTouch(
     view: JavaObject?,
     touchParameters: JavaObject
 ) {
-    if env == nil || view == nil {
-        // We're in an invalid state where the JNI has exploded somehow
-        // Passing anything on to UIKit now would probably lead to a crash
-        return
-    }
```
ephemer, line 243: We added this to try to fix the crash, but adding it didn't actually improve anything and I don't believe this was ever the actual issue. Since the stack trace indicates that the crash indeed currently occurs when creating the SDL event, and not before it, we will see if we get crashes _not_ from the async `Task` in this function, which might indicate an issue with the jni functions here after all. I suspect that won't be the case.

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -250,27 +244,29 @@ public func onNativeTouch(
     let pressure: JavaFloat = try! jni.GetField("pressure", from: touchParameters)
     let timestampMs: JavaLong = try! jni.GetField("timestamp", from: touchParameters)
```
ephemer, line 245: FYI @michaelknoch if the current fix does work, then we can probably revert this code back to how it was as well (i.e. that Kotlin just sends these parameters directly, rather than sending an object that contains them)

Sources/UIApplication+handleSDLEvents.swift
ephemer: Update: no, actually this is almost certainly the crash. 49 days of system uptime would lead to a UInt32 overflowing. It's not actually that unlikely that a device has been on for 49 days, especially newer, better devices. I will see if I can improve this further

ephemer: I further improved this by making `timestampMs` start from the time the SDLActivity was inited, rather than from when the device was started. This combined with `truncatingIfNeeded` should fix this problem permanently and without weirdness 🙏🏼

michaelknoch (approved):

PR #384, opened 2022-12-09
Build via Swift package manager

ephemer:
**Type of change:** DX, Performance

Re-allow builds via SwiftPM, which should make it easier for others to integrate the library into their own builds. We also gain improved support for static linking, whole module optimization, etc. by doing this.


Sources/Math.swift
```diff
@@ -0,0 +1,25 @@
+#if os(WASI)
```
michaelknoch, line 1: nit: WASI shouldn't be needed for UIKit, right?

Sources/Logging.swift
```diff
@@ -6,13 +7,8 @@
 //  Copyright © 2017 flowkey. All rights reserved.
 //
 
-import SDL
-
-#if os(Android)
+@_implementationOnly import SDL
```
michaelknoch, line 10: I found some docs about this [here](https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#_implementationonly) but can you explain in our context what this actually changes?

Sources/UIScrollView+velocity.swift
```diff
@@ -34,7 +34,7 @@ extension UIScrollView {
         let animationTimeConstant = 0.325 * dampingFactor
 
         // This calculation is a weird approximation but it's close enough for now...
-        let animationTime = log(Double(distanceToBoundsCheckedTarget.magnitude)) * animationTimeConstant
+        let animationTime = log2(Double(distanceToBoundsCheckedTarget.magnitude)) * animationTimeConstant
```
michaelknoch, line 37: is this change intended?

Sources/UIScrollView.swift
```diff
@@ -11,9 +11,10 @@
 let UIScrollViewDecelerationRateNormal: CGFloat = 0.998
 let UIScrollViewDecelerationRateFast: CGFloat = 0.99
 
+@MainActor
 open class UIScrollView: UIView {
     open weak var delegate: UIScrollViewDelegate? // TODO: change this to individually settable callbacks
-    open var panGestureRecognizer = UIPanGestureRecognizer()
+    open var panGestureRecognizer: UIPanGestureRecognizer
```
michaelknoch, line 17: is it not possible anymore to init here in that line when using @MainActor?

src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -36,10 +36,6 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
         // and it is a major overhaul to change the native C code (there are a lot of dependencies)
         @JvmStatic
         internal var mSeparateMouseAndTouch = false
-
-        init {
-            arrayOf("JNI", "SDL", "UIKit").forEach { System.loadLibrary(it) }
-        }
```
michaelknoch, line 42: do we use static linking already?

michaelknoch (approved):
nice stuff! Looks good to me, I left a few comments where things were unclear to me

rikner (approved):

rikner (approved):
@ephemer just realized we haven't merged this yet. Approving this again 🚀

PR #385, opened 2023-03-17
Implement UserDefaults.standard for Android

rikner:
Fixes # 

**Type of change:** 

Be able to permanently save primitives to the apps storage on the device across app sessions.

### DIFF
--- a/Sources/UserDefaults.swift
+++ b/Sources/UserDefaults.swift
@@ -1,0 +1,66 @@
+#if os(Android)
+
+import JNI
+
+
+private let TAG = "UserDefaults"
+
+public class UserDefaults: JNIObject {
+    override public static var className: String {
+        return "com.flowkey.uikit.UserDefaults"
+    }
+
+    public static var standard = try! UserDefaults()
+
+    private convenience init() throws {
+        let context = try jni.call("getContext", on: getSDLView(), returningObjectType: "android.content.Context")
+        try self.init(arguments: JavaContext(context))
+    }
+
+    public func has(_ itemKey: String) -> Bool {
+        return try! UserDefaults.standard.call(
+            methodName: "has",
+            arguments: [itemKey]
+        )
+    }
+
+    public func get(_ itemKey: String) -> Int {
+        let defaultValue = 0
+        return try! UserDefaults.standard.call(
+            methodName: "getInt",
+            arguments: [itemKey, defaultValue]
+        )
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/com.flowkey.uikit/UserDefaults.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun getInt(key: String, defValue: Int): Int {
+       return preferences.getInt(key, defValue)
+    }
+
+    fun getBoolean(key: String, defValue: Boolean): Boolean {
+        return preferences.getBoolean(key, defValue)
+    }
+
+    fun getString(key: String, defValue: String): String {
+        return preferences.getString(key, defValue) ?: defValue
```
rikner, line 25: This written a bit weird to make the compiler happy. It requires us to provide a default value but infers the return type of `preferences.getString()` to be `String?` :trollface:

Sources/UserDefaults.swift
```diff
@@ -0,0 +1,66 @@
+#if os(Android)
+
+import JNI
+
+
+private let TAG = "UserDefaults"
```
michaelknoch, line 6: is this variable used from somewhere?

Sources/UserDefaults.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public static var standard = try! UserDefaults()
+
+    private convenience init() throws {
+        let context = try jni.call("getContext", on: getSDLView(), returningObjectType: "android.content.Context")
+        try self.init(arguments: JavaContext(context))
+    }
+
+    public func has(_ itemKey: String) -> Bool {
```
michaelknoch, line 20: `has` doesn't seem to exist in the original UserDefaults API, no?

Sources/UserDefaults.swift
```diff
@@ -0,0 +1,66 @@
...
+            methodName: "has",
+            arguments: [itemKey]
+        )
+    }
+
+    public func get(_ itemKey: String) -> Int {
+        let defaultValue = 0
+        return try! UserDefaults.standard.call(
+            methodName: "getInt",
+            arguments: [itemKey, defaultValue]
```
michaelknoch, line 31: I do not fully get why we pass in defaults. Can't we return optionals?

rikner:
closing this as we rather might want to wait for Swift-only implementation of Foundation instead.

PR #390, opened 2024-01-23
update sdks of sample apps, fix non null assertion in kotlin code

michaelknoch:
**Type of change:** bugfix

Fixes a compilation assertion in SDLActivity

```
> Task :UIKit:compileDebugKotlin FAILED
e: file:///mobile-app/RNPlayerComponent/NativePlayer/UIKit/src/main/java/org/libsdl/app/SDLActivity.kt:241:30 Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type InputDevice?
e: file:///mobile-app/RNPlayerComponent/NativePlayer/UIKit/src/main/java/org/libsdl/app/SDLActivity.kt:241:73 Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type InputDevice?
```

### DIFF
--- a/src/main/java/org/libsdl/app/SDLActivity.kt
+++ b/src/main/java/org/libsdl/app/SDLActivity.kt
@@ -237,7 +237,7 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
     @Suppress("unused")
     fun inputGetInputDeviceIds(sources: Int): IntArray {
         return InputDevice.getDeviceIds().fold(intArrayOf(0)) { result: IntArray, id: Int ->
-            val device = InputDevice.getDevice(id)
+            val device = InputDevice.getDevice(id) ?: return result
             return if (device.sources and sources != 0) (result + device.id) else result
         }
     }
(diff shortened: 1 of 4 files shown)


src/main/java/org/libsdl/app/SDLActivity.kt
```diff
@@ -237,7 +237,7 @@ open class SDLActivity internal constructor (context: Context?) : RelativeLayout
     @Suppress("unused")
     fun inputGetInputDeviceIds(sources: Int): IntArray {
         return InputDevice.getDeviceIds().fold(intArrayOf(0)) { result: IntArray, id: Int ->
-            val device = InputDevice.getDevice(id)
+            val device = InputDevice.getDevice(id) ?: return result
```
michaelknoch, line 240: the weird thing is that this is not causing an issue when building the sample app or the standalone player. I only run into this when building the app. Kotlinversion and sdks were all aligned, still no compilation error. The new variant is also building from the sample app and the player though.

rikner (approved):

PR #393, opened 2024-05-14
Improve UIView.safeAreaInsets

rikner:
Fixes # 

**Type of change:** 

Make `UIView.safeAreaInsets` a computed property which always returns the current insets  from the Android API. This will reflect changes when the device orientation changes.

### DIFF
--- a/Sources/UIView.swift
+++ b/Sources/UIView.swift
@@ -43,7 +43,9 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
         }
     }
 
-    open internal(set) var safeAreaInsets: UIEdgeInsets = .zero
+    public var safeAreaInsets: UIEdgeInsets {
+        return UIWindow.getSafeAreaInsets()
+    }
 
     open var mask: UIView? {
         didSet {
--- a/Sources/UIWindow.swift
+++ b/Sources/UIWindow.swift
@@ -19,7 +19,6 @@ public class UIWindow: UIView {
     }
 
     open func makeKeyAndVisible() {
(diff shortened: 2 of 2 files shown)


michaelknoch (approved):

Sources/UIView.swift
```diff
@@ -43,7 +43,9 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
         }
     }
 
-    open internal(set) var safeAreaInsets: UIEdgeInsets = .zero
+    public var safeAreaInsets: UIEdgeInsets {
+        return UIWindow.getSafeAreaInsets()
```
ephemer, line 47: This will solve the problem with insets not updating on orientation change but it will also cause unneeded traffic via the JNI. It would be better if we could somehow push these changes down to the views when the orientation changes.

Also it's not actually correct that the safe area insets of every view are equal to those of the UIWindow – the insets will vary depending on the view's position inside the window (if you have a 50x50 rectangle in the centre of the screen, then that should have no insets. If that same rectangle is placed at an edge of the screen where UIWindow has safe area insets, it should adopt those insets on the appropriate side)

For now I think it's ok as is but both of those things make me a little uneasy about keeping this solution for the longer term

rikner: Agree. We also had some doubts about this for the reasons you mentioned but thought it's does the job for now. Good to know you have the same thoughts  👍 Let's leave it on our radar for now and maybe we can tackle this in two steps

a) calculate the correct safe area insets depending on the views frame/bounds
b) push these changes down to the views when the orientation changes instead of polling them via the JNI

PR #395, opened 2024-07-31
add new initializer for transform rotation to CGAffineTransform extension

michaelknoch:
**Type of change:** feature

Add `public init(rotationAngle: CGFloat)`


reckjn (approved):

Sources/CGAffineTransform.swift
```diff
@@ -13,8 +13,11 @@ public extension CGAffineTransform {
         self.init(scaleByX: scaleX, byY: y)
     }
 
+    init(rotationAngle: CGFloat) {
+        self.init(rotationByRadians: rotationAngle)
+    }
```
ephemer, line 18: I was a bit confused by this and just looked into it. Apparently `AffineTransform` is a Foundation (only?) type that has this `rotationByRadians` initializer. CGAffineTransform appears to have different initializers.

Our CGAffineTransform type is copied from Foundation's AffineTransform type, so it's a bit of a weird mix of both. I'm not sure what implications this has for code re-use. It's probably "close-enough" API-wise that we can use it as is, especially given this new extension. But we should be aware of the differences

@michaelknoch @reckjn

michaelknoch: Got it, thanks for the info. I was confused as well tbh. Do you think our CGAffineTransform needs to be refactored because of this?

PR #397, opened 2024-08-07
cleanup API of AffineTransform

michaelknoch:
**Type of change:** cleanup

Based on the conversations in https://github.com/flowkey/UIKit-cross-platform/pull/395#discussion_r1707493648

### DIFF
--- a/Sources/AffineTransform.swift
+++ b/Sources/AffineTransform.swift
@@ -110,7 +110,7 @@ extension AffineTransform {
     /// [ -sin α   cos α  0 ]
     /// [    0       0    1 ]
     /// ```
-    public init(rotationByRadians angle: CGFloat) {
+    public init(rotationAngle angle: CGFloat) {
         let sinα = sin(angle)
         let cosα = cos(angle)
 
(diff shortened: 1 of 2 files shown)


ephemer (approved):
Hi @michaelknoch, I am approving this because in a sense I do consider it an improvement (assuming that the type we're actually intending to vend is `CGAffineTransform` and not `AffineTransform`).

While reviewing this PR though I can see that there would still be quite a bit of cleanup to do in order to "do it properly", i.e. actually renaming `AffineTransform` to CGAffineTransform, removing the other initializers that don't exist in CGAffineTransform (like `rotationByDegrees`), and moving the initializers that are currently in the CGAffineTransform.swift extension into the type itself (and removing their AffineTransform aliases, e.g. `scaleByX...` as well).

The above doesn't really need to happen before merging this, again assuming that `CGAffineTransform` is the type we actually intend to vend here. But to "do it properly" there would still be some work here to be done.

I can see that I made `CGAffineTransform` in May 2017, and from memory I also started working on this project – at least full time – in May 2017 too. So this must be some of the earliest code in the project 😵

PR #398, opened 2024-08-26
improve absolute touch handling

michaelknoch:
**Type of change:** bugfix

improve absolute touch handling

### DIFF
--- a/Sources/UIScreen.swift
+++ b/Sources/UIScreen.swift
@@ -59,7 +59,7 @@ public final class UIScreen {
         var size = CGSize.zero
         let options: SDLWindowFlags = [SDL_WINDOW_FULLSCREEN]
         #else
-        var size = CGSize.samsungGalaxyTab10.portrait
+        var size = CGSize.samsungGalaxyJ5.landscape
 
         let options: SDLWindowFlags = [
             SDL_WINDOW_ALLOW_HIGHDPI,
--- a/Sources/UIScreen.swift
+++ b/Sources/UIScreen.swift
@@ -59,7 +59,7 @@ public final class UIScreen {
         var size = CGSize.zero
         let options: SDLWindowFlags = [SDL_WINDOW_FULLSCREEN]
         #else
-        var size = CGSize.samsungGalaxyJ5.landscape
+        var size = CGSize.samsungGalaxyTab10.portrait
 
         let options: SDLWindowFlags = [
             SDL_WINDOW_ALLOW_HIGHDPI,
@@ -192,7 +192,7 @@ fileprivate func getAndroidDeviceScale() -> CGFloat {
 extension UIScreen {
     /// Used in tests only and doesn't actually render anything
     static func dummyScreen(
-        bounds: CGRect = CGRect(origin: .zero, size: .samsungGalaxyS7),
+        bounds: CGRect = CGRect(origin: .zero, size: .samsungGalaxyTab10),
         scale: CGFloat
     ) -> UIScreen {
         return UIScreen(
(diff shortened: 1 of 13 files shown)


Sources/UIScrollView.swift
```diff
@@ -54,13 +54,22 @@ open class UIScrollView: UIView {
...
         return CGPoint(
-            x: min(max(newContentOffset.x, -contentInset.left), (contentWidth + contentInset.right) - bounds.width),
-            y: min(max(newContentOffset.y, -contentInset.top), (contentHeight + contentInset.bottom) - bounds.height)
+            // avoid unexpected negative offset if content is smaller than available size
+            x: contentSize.width <= bounds.width
+                ? minXOffset
+                : max(minXOffset, min(newContentOffset.x, maxXOffset)),
+            y: contentSize.height <= bounds.height
+                ? minYOffset
+                : max(minYOffset, min(newContentOffset.y, maxYOffset))
```
michaelknoch, line 72: fyi @rikner : these changes should fix the horizontal scrolling full sheet

UIKitTests/UIView/UIViewTests+convert.swift
```diff
@@ -94,7 +94,7 @@ class UIViewPointConversionTests: XCTestCase {
         rootView.addSubview(subview1)
         subview1.addSubview(subview1subview1)
 
-        XCTAssertEqual(subview1subview1.absoluteOrigin(), CGPoint(x: 15, y: 15))
```
michaelknoch, line 97: these changes here might look random, but we got rid of our custom implementation of the `absoluteOrigin` function for iOS. This now uses UIView convert internally, which should be more correct

Sources/UIView.swift
```diff
@@ -258,7 +258,7 @@ open class UIView: UIResponder, CALayerDelegate, UIAccessibilityIdentification {
         let otherAbsoluteOrigin = otherView.absoluteOrigin()
 
         let originDifference = (otherAbsoluteOrigin - selfAbsoluteOrigin)
-        return point - originDifference
+        return (point - originDifference).applying(otherView.superview?.transform.inverted() ?? .identity)
```
michaelknoch, line 261: This code was moved from `Sources/UIPanGestureRecognizer.swift`. It's still a hack, in my current understanding we have to consider all transforms between self and otherView (which we partially do in `absoluteOrigin` 🤯 )

Sources/UIScreen.swift
```diff
@@ -59,7 +59,7 @@ public final class UIScreen {
         var size = CGSize.zero
         let options: SDLWindowFlags = [SDL_WINDOW_FULLSCREEN]
         #else
-        var size = CGSize.samsungGalaxyJ5.landscape
+        var size = CGSize.samsungGalaxyTab10.portrait
```
ephemer, line 62: would it make sense to keep this default how it was?

Sources/UIView.swift
ephemer: It would be great if you could add a code comment that this is still kind of a hack and ideally it would incorporate all transforms up the tree

UIKitTests/UIView/UIViewTests+convert.swift
```diff
@@ -243,6 +243,36 @@ class UIViewPointConversionTests: XCTestCase {
         XCTAssertEqual(convertedPoint1, convertedPoint2)
         XCTAssertEqual(convertedPoint1, CGPoint(x: 50, y: 50))
     }
+
+    func testConversionWithMultipleSubviewsBounds() {
```
ephemer, line 247: It would be great if you could add the `transform`s back in here but commented out (we could even keep of them for now I think). It would be really nice to have a clear place to work on when we come back to this someday

ephemer (approved):
Looks good to me! My main comments here are requests to make it clear what to work on when we come back to this some day.

I would consider changing the PR title to "improve absolute touch handling" because it's not _quite_ fixed in this PR yet

ephemer:
Thanks a lot for diving back into this crazy part of the code @michaelknoch <3

rikner (approved):
💯

PR #399, opened 2024-08-28
implement proper interpolation of CATransform3D for animations

michaelknoch:
**Type of change:** feature

This change improves the interpolation of CATransform3D by individually interpolating its components. Additionally, it utilizes spherical linear interpolation (SLERP) to handle rotation animations, ensuring more accurate rotation transitions.

**Before:**
Aug-28-2024 09-54-36

**After:**
Aug-28-2024 09-54-40


Sources/CATransform3D+interpolate.swift
```diff
@@ -0,0 +1,237 @@
...
+private func sqrt(_ value: Float) -> Float {
+    return Float(sqrt(Double(value)))
+}
+private func acos(_ value: Float) -> Float {
+    return Float(acos(Double(value)))
+}
+private func sin(_ value: Float) -> Float {
+    return Float(sin(Double(value)))
+}
+#endif
```
michaelknoch, line 237: These functions expect doubles on android. Is there a better workaround than this? The type conversions might cause some minor overhead

rikner: aren't there float equivalents `sqrtf`, `acosf` and `sinf`?

michaelknoch: Nice thanks, I didn't know these exist. I cleaned this up in https://github.com/flowkey/UIKit-cross-platform/pull/399/commits/ced4e407a48f91d1992f0207b632e40d7e1a2e66

UIKit.xcodeproj/project.pbxproj
```diff
@@ -1003,6 +1014,13 @@
 /* End PBXProject section */
 
 /* Begin PBXReferenceProxy section */
+		5BF8AFEA2C7E95C200065D8E /* lolTests.xctest */ = {
```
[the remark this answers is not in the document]
michaelknoch: ups, remove it in 7bb2452935c521dc77c276bd371ab5e11ead13a9

rikner (approved):
Not much clue about the math behind this, but the code looks really clean 💯

Sources/CATransform3D+interpolate.swift
ephemer: the float equivalents are C functions that are available in `import Glibc` (which is actually the "Bionic" library and not Glibc – Android is a mess). They are also available on Apple platforms if you `import Darwin`.

For squareRoot specifically I would recommend against using free-standing functions and use the versions from the Swift standard library (`myFloat.squareRoot()`). The trigonometry functions don't exist in that form unfortunately so still need to be imported from somewhere

ephemer: The long term solution to this is to use `import Numerics`, which is available as a Swift package. It also introduces more performance and correctness fixes (including for trig functions)

https://www.swift.org/blog/numerics/

ephemer (approved):
Overall this looks good to me, thanks for the PR @michaelknoch! We can optimize the interpolation and other code by using the Swift Numerics package, doing things like the 


```
// If divisor is well-scaled, use multiply by reciprocal instead of division.
if let recip = divisor.reciprocal {
  return data.map { $0 * recip }
}
// Otherwise, fallback on using division.
return data.map { $0 / divisor }
```

example when performing many divisions by the same divisor. It might also make our code a bit more portable due to Numerics handling the definition or import of various underlying math functions.

https://www.swift.org/blog/numerics/

This can happen in a future PR though – it should not be blocking for this one. I'm pretty confident the performance with this change should be good enough for now

Sources/CATransform3D+interpolate.swift
michaelknoch: cool, thanks for the feedback. I changed the lines where I used `sqrtf()` to use `. squareRoot()`. I'll look into using the Numerics package in a separate iteration. It looks interesting, thanks for the hint!

PR #401, opened 2024-10-18
Get initial dimensions for UIScreen via JNI rather than via `SDLDisplayMode.current`

michaelknoch:
**Type of change:** bugfix

We get the screen dimensions from `SDLDisplayMode.current` for initializing UIscreen. There is a bug currently where the dimensions can be wrong (flipped) when an orientation change has happened before re-initalizing UIscreen. See steps to reproduce for more info.

#### Steps to reproduce:
- Init UIKit App (init SDLActivity, UIApplication, UIScreen etc.)
- Change orientation in UIKit app
- Leave UIKit app -> deinit everything, call `SDL_quit`
- Change orientation outside of UIKit App
- Re-init UIKit App

[Screen_recording_20241018_131951.webm](https://github.com/user-attachments/assets/535dd165-e7ff-4bdb-8873-c36a2270ed99)


Sources/UIScreen.swift
```diff
@@ -70,12 +70,20 @@ public final class UIScreen {
...
+
         if options.contains(SDL_WINDOW_FULLSCREEN), let displayMode = SDLDisplayMode.current {
             // Fix fullscreen resolution on Mac and make Android easier to reason about:
             // There is an inconsistency between Mac and Android when setting SDL_WINDOW_FULLSCREEN
             // The easiest solution is just to work in 1:1 pixels
             GPU_SetPreInitFlags(GPU_GetPreInitFlags() | GPU_INIT_DISABLE_AUTO_VIRTUAL_RESOLUTION)
+            #if os(Android)
+            size = CGSize(width: CGFloat(androidScreenDimension.width), height: CGFloat(androidScreenDimension.height))
+            #else
             size = CGSize(width: CGFloat(displayMode.w), height: CGFloat(displayMode.h))
```
michaelknoch, line 85: these dimensions here from `SDLDisplayMode.current` are wrong in that specific case. It seems that the displayMode caches the previous dimension and I did not find a way to reset it. We do properly deinit UIScreen, which calls GPU_quit

ephemer: I thought that we set `SDLDisplayMode.current` ourselves somewhere. In that case would an option be to update it on orientation change rather than working around the issue?

ephemer (approved):
It looks good to me, other than it adding code paths that may not be needed if we can update the displaymode correctly?

Sources/UIScreen.swift
ephemer: By "we" I mean some code that we have control over, which is "everything", for better or worse 😅 I am pretty sure we could fix this at the source, but this workaround looks fine to me

michaelknoch: Agreed!

PR #402, opened 2024-10-30
Swift 6 support

ephemer:
Fixes # 

**Type of change:** 

### DIFF
--- a/Sources/AVPlayer+Android.swift
+++ b/Sources/AVPlayer+Android.swift
@@ -6,7 +6,9 @@
 //  Copyright © 2017 flowkey. All rights reserved.
 //
 
+#if os(Android)
 import JNI
+#endif
 
 public class AVPlayer: JNIObject {
     public override static var className: String { "org.uikit.AVPlayer" }
--- a/Sources/AVPlayer+Android.swift
+++ b/Sources/AVPlayer+Android.swift
@@ -6,9 +6,7 @@
 //  Copyright © 2017 flowkey. All rights reserved.
 //
 
-#if os(Android)
 import JNI
-#endif
 
 public class AVPlayer: JNIObject {
     public override static var className: String { "org.uikit.AVPlayer" }
--- a/Sources/SDL+JNIExtensions.swift
+++ b/Sources/SDL+JNIExtensions.swift
@@ -6,9 +6,7 @@
 //  Copyright © 2017 flowkey. All rights reserved.
 //
 
-#if os(Android)
 import JNI
-#endif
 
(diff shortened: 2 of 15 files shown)


Sources/AVPlayer+Android.swift
```diff
@@ -6,7 +6,9 @@
 //  Copyright © 2017 flowkey. All rights reserved.
 //
 
+#if os(Android)
```
ephemer, line 9: I think this entire file should be Android only, no? It would not work anywhere else

Sources/SDL+JNIExtensions.swift
```diff
@@ -6,7 +6,9 @@
 //  Copyright © 2017 flowkey. All rights reserved.
 //
 
+#if os(Android)
 import JNI
+#endif
```
ephemer, line 11: this file would also only work on Android, so can we get away with excluding this file from the build entirely rather than just adding this conditional on the import?

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -250,8 +253,10 @@ extension UITouch: CustomStringConvertible {
     }
 }
 
+#if os(Android)
 #if os(Android)
 import JNI
+#endif
```
ephemer, line 259: this needs some attention

Sources/UIApplication+handleSDLEvents.swift
```diff
@@ -7,7 +7,10 @@
 //
 
 import SDL
+
+#if os(Android)
 import JNI
+#endif
```
ephemer, line 13: I think we can remove this since we're already importing JNI in the same file below?

Sources/UIApplication.swift
```diff
@@ -77,8 +77,10 @@ public enum UIStatusBarStyle {
 }
 
 
+#if os(Android)
 #if os(Android)
 import JNI
+#endif
```
ephemer, line 83: ?

Sources/UIScreen.swift
```diff
@@ -183,8 +183,10 @@ extension UIScreen {
 }
 #endif
 
+#if os(Android)
 #if os(Android)
 import JNI
+#endif
```
ephemer, line 189: ?

Sources/androidNativeInit.swift
```diff
@@ -7,7 +7,9 @@
 //
 
 import SDL
+#if os(Android)
 import JNI
+#endif
```
ephemer, line 12: this entire file should be Android only AFAIK

rikner (approved):
👍

PR #406, opened 2025-10-10
Allow multiple AVPlayer instances, use TextureView instead of SurfaceView

michaelknoch:
**Type of change:** feature

Allow multiple AVPlayer instances, use TextureView instead of SurfaceView

### DIFF
--- a/Sources/CALayer.swift
+++ b/Sources/CALayer.swift
@@ -242,6 +242,14 @@ open class CALayer {
         didSet { onDidSetAnimations(wasEmpty: oldValue.isEmpty) }
     }
 
+    open func animationKeys() -> [String]? {
+        return animations.keys.isEmpty ? nil : animations.keys.map { $0 }
+    }
+
+    open func animation(forKey key: String) -> CABasicAnimation? {
+        return animations[key]
+    }
+
     /// We disable animation on parameters of views / layers that haven't been rendered yet.
     /// This is both a performance optimization (avoids lots of animations at the start)
     /// as well as a correctness fix (matches iOS behaviour). Maybe there's a better way though?
--- a/Sources/CALayer.swift
+++ b/Sources/CALayer.swift
@@ -182,7 +182,10 @@ open class CALayer {
     public required init() {}
 
     public required init(layer: Any) {
-        guard let layer = layer as? CALayer else { fatalError() }
+        guard let layer = layer as? CALayer else {
+            fatalError("Copy of CALayer must be initialized from another CALayer")
+        }
+
         bounds = layer.bounds
         delegate = layer.delegate
         transform = layer.transform
@@ -208,8 +211,8 @@ open class CALayer {
         contentsGravity = layer.contentsGravity
     }
 
-    open func copy() -> Any {
-        return CALayer(layer: self)
+    open func copy() -> Self {
+        return Self(layer: self)
     }
 
     open func action(forKey event: String) -> CAAction? {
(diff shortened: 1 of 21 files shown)


Sources/CALayer.swift
```diff
@@ -221,7 +224,11 @@ open class CALayer {
 
     /// returns a non animating copy of the layer
     func createPresentation() -> CALayer {
-        let copy = CALayer(layer: self)
+        // XXX: Should we just return _presentation if it already exists??
+        // This seems to break animations, but why?
+        // if let _presentation { return _presentation }
```
ephemer, line 229: @michaelknoch this is kind of an open question to you I think. We don't need to answer it right now but I it did confuse me that we create a new presentation layer on every frame – is that how iOS works?

Sources/CGImage.swift
```diff
@@ -43,12 +43,34 @@ public class CGImage {
...
+            var height: Int32 = 0
+            var channels: Int32 = 4
+
+            #if os(Android)
+            // Android natively supports 2-channel textures. Use them to save 50% (GPU) RAM.
+            let data = stbi_load_from_memory(buffer.baseAddress, Int32(buffer.count), &width, &height, &channels, 0)
+
+            let format: GPU_FormatEnum = switch channels {
+            case 1: GPU_FORMAT_ALPHA
+            case 2: GPU_FORMAT_LUMINANCE_ALPHA
```
ephemer, line 56: @michaelknoch FYI in theory GPU_FORMAT_LUMINANCE_ALPHA (and GPU_FORMAT_ALPHA) is not supported in OpenGLES in versions 3+. So this "should" not work on newer devices. In practice I saw that Android emulates OpenGLES2 in these cases, so it works. I don't think it's problematic, but I wanted to mention it.

I didn't immediately find a way to see if this device is running OpenGLES2 vs OpenGLES3 – if we can find that we could also fall back to using 4-channel textures in that case. But let's see if it's actually a problem first

Sources/CGImage.swift
```diff
@@ -43,12 +43,34 @@ public class CGImage {
...
+            case 3: GPU_FORMAT_RGB
+            case 4: GPU_FORMAT_RGBA
+            default: fatalError()
             }
-
-            let rw = SDL_RWFromMem(ptr, Int32(buffer.count))
-            return GPU_LoadImage_RW(rw, true)
+            #elseif os(macOS)
```
ephemer, line 61: i'm not sure we need `elseif os(macOS)` here – it should be just "else"

src/main/java/org/uikit/VideoJNI.kt
```diff
@@ -156,28 +163,44 @@ class AVPlayer(parent: SDLActivity, asset: AVURLAsset) {
...
         }
     }
 
-    fun setResizeMode(resizeMode: Int) {
-        exoPlayerView.resizeMode = resizeMode
+    fun setIsHidden(newValue: Boolean) {
+        // `newValue` is the HIDDEN state, whereas we're setting the _VISIBILITY_ here:
+        visibility = if (!newValue) { View.VISIBLE } else { View.INVISIBLE }
+        // Note: there is another visibility state, `View.GONE`, which also removes the view
+        // from layout (similar to `display: none`), but we want to match iOS behaviour here.
```
ephemer, line 297: @michaelknoch actually i'm unsure if GONE isn't a closer match to iOS behaviour. We kind of don't care about the layouting either way though because essentially our views are positioned absolutely. Either way, GONE might actually make more sense? Do you have any thoughts about it?

ephemer (approved):
It looks good to me (I hope so because I think I wrote most of it)

PR #408, opened 2025-12-12
Add CAGradientLayer, Allow Layer Flattening for proper masking and opacity compositing

ephemer:
**Type of change:** Feature

- Add a simple CPU-only (for now) CAGradientLayer and the beginnings of a GPU implementation.
- Add layer flattening
    - With this, non-trivial layers with non-zero opacity and/or a mask are rendered as a separate render tree and then composited onto the window (or into each other)
    - This enables proper masking, e.g. via a gradient layer (via alpha) or using non-trivial shapes


Sources/AVPlayerLayer+Mac.swift
```diff
@@ -109,16 +108,7 @@ public final class AVPlayerLayer: CALayer {
...
-        }
-
-        // Swap R and B values to get RGBA pixels instead of BGRA:
-        let bufferSize = CVPixelBufferGetDataSize(pixelBuffer)
-        for i in stride(from: 0, to: bufferSize, by: 16) {
-            swap(&pixelBytes[i], &pixelBytes[i + 2])
-            swap(&pixelBytes[i+4], &pixelBytes[i + 6])
-            swap(&pixelBytes[i+8], &pixelBytes[i + 10])
-            swap(&pixelBytes[i+12], &pixelBytes[i + 14])
+            contents = VideoTexture(width: width, height: height, format: GPU_FORMAT_BGRA)
```
ephemer, line 111: this change dramatically improves video performance on macOS – esp. in debug builds

Sources/CALayer+ContentsGravity.swift
```diff
@@ -22,8 +22,8 @@ struct ContentsGravityTransformation {
     /// Warning, this assumes `layer` has `contents` and will crash otherwise!
     init(for layer: CALayer) {
         let scaledContents = CGSize(
-            width: CGFloat(layer.contents!.width) / layer.contentsScale,
-            height: CGFloat(layer.contents!.height) / layer.contentsScale
+            width: (layer.contents.map { CGFloat($0.width) } ?? layer.bounds.width) / layer.contentsScale,
+            height: (layer.contents.map { CGFloat($0.height) } ?? layer.bounds.height) / layer.contentsScale
```
ephemer, line 26: turns out this is actually unused but it prevents the crash mentioned on line 22 (I could remove that comment now I guess)

Sources/CALayer+SDL.swift
```diff
@@ -1,17 +1,9 @@
...
-//  Copyright © 2017 flowkey. All rights reserved.
-//
-
 internal import SDL_gpu
 
 extension CALayer {
     @MainActor
-    final func sdlRender(parentAbsoluteOpacity: Float = 1) {
-        guard let renderer = UIScreen.main else { return }
+    final func sdlRender(parentAbsoluteOpacity: Float = 1, renderTarget: RenderTarget) {
```
ephemer, line 5: the concept of the RenderTarget being able to change is the main innovation in this PR: to begin with we start rendering to the window (which is the base-level RenderTarget). If we encounter a non-trivial CALayer with a mask/opacity, we render its entire subtree to a separate RenderTarget (which is literally an image in memory), and then we copy that pre-rendered image onto the window with the correct mask and opacity.

Sources/CGImage.swift
```diff
@@ -39,6 +39,10 @@ public class CGImage {
         height = Int(rawPointer.pointee.h)
     }
 
+    public func savePNG(filename: String) {
```
ephemer, line 42: debugging API – I should probably prefix this with an underscore

Sources/FontRenderer+singleLineSize.swift
```diff
@@ -114,7 +114,7 @@ extension FontRenderer {
             let spaceCharacterCode = 32
             let newLineCharacterCode = 10
             if characterCode != spaceCharacterCode, characterCode != newLineCharacterCode, glyph.maxx - glyph.minx <= 0 {
-                assertionFailure("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') has no width")
+//                assertionFailure("Glyph \(characterCode) ('\(Character(UnicodeScalar(characterCode)!))') has no width")
```
ephemer, line 117: this change was unintentional – will revert it

Sources/RenderTarget.swift
```diff
@@ -0,0 +1,192 @@
...
+
+    func flip() throws {
+        GPU_Flip(rawPointer)
+        try throwOnErrors(ofType: [GPU_ERROR_USER_ERROR, GPU_ERROR_BACKEND_ERROR])
+    }
+
+    func setVirtualResolution(w: UInt16, h: UInt16) {
+        GPU_SetVirtualResolution(rawPointer, w, h)
+    }
+}
```
ephemer, line 152: almost everything here is copied directly from UIScreen+render.swift -> previously UIScreen was our only render target, but now we can have various targets (see comments and code above)

Sources/ShaderProgram+Gradient.swift
```diff
@@ -0,0 +1,124 @@
+//
+//  Mask.swift
+//  UIKit
+//
+//  Created by Geordie Jay on 25.10.17.
+//  Copyright © 2017 flowkey. All rights reserved.
+//
```
ephemer, line 7: should probably remove this..

Sources/ShaderProgram+Gradient.swift
```diff
@@ -0,0 +1,124 @@
...
+
+final class GradientShaderProgram: ShaderProgram {
+    private var colorCount: UniformVariable!
+    private var gradientColors: UniformVariable!
+    private var gradientLocations: UniformVariable!
+
+    private var startPoint: UniformVariable!
+    private var endPoint: UniformVariable!
+
+    // we only need one MaskShaderProgram, which we can / should treat as a singleton
```
ephemer, line 23: remove comment

Sources/ShaderProgram+Gradient.swift
ephemer, line 1: This entire file is unused in this PR – it's a GPU implementation for `CAGradientLayer` but integrating it would require some more work, and we don't need the flexibility or the performance of using the GPU for our needs right now

Sources/UIView+SDL.swift
```diff
@@ -25,9 +25,13 @@ extension UIView {
             needsDisplay = false
         }
 
+        if visibleLayer.mask?.needsDisplay() == true {
+            visibleLayer.mask?.display()
+            visibleLayer.mask?._needsDisplay = false
+        }
```
ephemer, line 31: this can be removed because we (also) `display()` in the sdlRender function

michaelknoch (approved):

PR #410, opened 2026-05-21
Expose pixel-streaming primitives on CGImage 

michaelknoch:
**Type of change:** API

  Lets callers outside UIKit produce raw pixels and stream them into a GPU-backed layer texture
                                                                                                                                        
  - `CGImage(width:height:format:)` — allocate an empty GPU texture
  - `replacePixels(with:bytesPerPixel:)` — mutate it in place via `GPU_UpdateImageBytes`
  - `PixelFormat` — `.rgba` / `.bgra` (GLES on Android only reliably supports `.rgba`)                                                  
                                                                                                                                        
  Also makes `Timer` and `Data._fromPathCrossPlatform` public so the same callers can do elapsed-time accounting and bundle-resource    
  loading.                                                                                                                              
                                                                                                                                        
  Used by NativePlayer's new Android Lottie pipeline (rlottie → swizzle → upload).    






Sources/CGImage.swift
```diff
@@ -81,11 +81,39 @@ public class CGImage {
...
+    public func replacePixels(with bytes: UnsafePointer<UInt8>, bytesPerPixel: Int) {
         var rect = GPU_Rect(x: 0, y: 0, w: Float(rawPointer.pointee.w), h: Float(rawPointer.pointee.h))
         GPU_UpdateImageBytes(rawPointer, &rect, bytes, Int32(rawPointer.pointee.w) * Int32(bytesPerPixel))
     }
 
+    public enum PixelFormat {
+        /// 32-bit per pixel, byte order R,G,B,A. Reliably supported on all backends.
+        case rgba
+        /// 32-bit per pixel, byte order B,G,R,A. Not supported on Android GLES — falling
+        /// back to `.rgba` and swizzling client-side is safer.
```
ephemer: i'm not sure BGRA is truly unsupported. I think I recently got this working, but it may have been in a different context. Either way, ThorVG supports rendering as BGRA or RGBA as far as I know, so this may be unneeded there

Sources/CGImage.swift
```diff
@@ -81,11 +81,39 @@ public class CGImage {
         self.init(pointer, sourceData: nil)
     }
 
-    internal func replacePixels(with bytes: UnsafePointer<UInt8>, bytesPerPixel: Int) {
+    /// Allocates an empty GPU-backed image of the given size. Pixels are uninitialized —
+    /// callers must follow up with `replacePixels(with:bytesPerPixel:)`. Used when the
+    /// pixel content is produced per-frame by code outside UIKit (e.g. video decoders,
+    /// vector animation renderers).
+    public convenience init?(width: Int, height: Int, format: PixelFormat) {
```
ephemer: the reason i'm not a big fan of this is that these APIs don't exist in actual UIKit. I'm not too stoked about deviating significantly from what exists there

Sources/Data+fromRelativePathCrossPlatform.swift
```diff
@@ -9,7 +9,7 @@
 internal import SDL
 
 extension Data {
-    static func _fromPathCrossPlatform(_ path: String) -> Data? {
+    public static func _fromPathCrossPlatform(_ path: String) -> Data? {
```
ephemer: does this really need to be exposed? I thought Data had a natural `Data(from url: URL)` initializer? I remember there was a reason we created this workaround, but does it really need to be public?

Sources/Timer.swift
```diff
@@ -12,9 +12,9 @@ import Bionic
 import Darwin.C
 #endif
 
-struct Timer {
+public struct Timer {
```
ephemer: this doesn't exist in iOS so I'd rather not expose it. Can we avoid it?

michaelknoch:
nice! lgtm. Thanks @ephemer

PR #411, opened 2026-05-22
Fix 1-px edge noise on Android with fractional density

michaelknoch:
**Type of change:** bugfix

On devices reporting a non-integer density (e.g. Galaxy Tab S9 with `density=2.125`), a 1-px band of GPU garbage was visible at the right/bottom edges of the screen — sometimes random, sometimes leftover content from earlier frames. The noise got worse after orientation changes.

## Root cause

`UIScreen.bounds.width = physical_width / density` was fractional (e.g. `2560 / 2.125 = 1204.7`), but `GPU_SetVirtualResolution` only takes `UInt16` so `target->w` was rounded to `1205`. SDL-gpu's orthographic projection maps virtual `x = target->w` to NDC `+1`. Drawing a view at the *fractional* `bounds.width` landed at NDC ≈ `0.9994`, missing the rightmost physical pixel column entirely — those pixels never got cleared and never got drawn, so the SurfaceView composited the underlying framebuffer through them.

A secondary problem: `gpuRect` used a single `UIScreen.scale` to snap to physical pixel boundaries, but `drawable_w/target->w` and `drawable_h/target->h` are subtly different on non-integer densities. Using the x-ratio for y snapped the bottom edge ~0.2 virtual-px short of the viewport.

## Fix

- **`Sources/UIScreen.swift`** — snap `bounds.size` to the rounded virtual resolution and back-derive `scale` from the rounded value, so `bounds.width * scale == physical_width` *exactly*.
- **`Sources/UIScreen+render.swift`** — `gpuRect` now reads per-axis ratios directly from the GPU target (`drawable_w/target->w`, `drawable_h/target->h`) instead of a single `UIScreen.scale`.

## Screenshots

https://github.com/user-attachments/assets/f88f0990-32db-4c30-a1fa-70676ee3f589

IMG_7536

### DIFF
--- a/Sources/UIScreen+render.swift
+++ b/Sources/UIScreen+render.swift
@@ -84,9 +84,9 @@ extension UIScreen {
 
     func fill(_ rect: CGRect, with color: UIColor, cornerRadius: CGFloat) {
         if cornerRadius >= 1 {
-            GPU_RectangleRoundFilled(rawPointer, rect.gpuRect(scale: scale), cornerRadius: Float(cornerRadius), color: color.sdlColor)
+            GPU_RectangleRoundFilled(rawPointer, rect.gpuRect(target: rawPointer), cornerRadius: Float(cornerRadius), color: color.sdlColor)
         } else {
-            GPU_RectangleFilled(rawPointer, rect.gpuRect(scale: scale), color: color.sdlColor)
+            GPU_RectangleFilled(rawPointer, rect.gpuRect(target: rawPointer), color: color.sdlColor)
         }
     }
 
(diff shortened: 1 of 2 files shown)


mouselangelo (approved):
LGTM 🙏

AKiranB (approved):
LGTM

Sources/UIScreen+render.swift
```diff
@@ -136,17 +136,22 @@ extension UIScreen {
...
+        GPU_SetClipRect(rawPointer, clippingRect.gpuRect(target: rawPointer))
     }
 }
 
 private extension CGRect {
-    func gpuRect(scale: CGFloat) -> GPU_Rect {
+    // Snap to physical pixel boundaries using the GPU target's per-axis ratios —
+    // a single `UIScreen.scale` is wrong when drawable_w/target_w ≠ drawable_h/target_h
+    // (any fractional density rounds the two axes by different amounts).
+    func gpuRect(target: UnsafeMutablePointer<GPU_Target>!) -> GPU_Rect {
```
ephemer: Since this is a private extension, it seems like it might make more sense for this to "live" on UIScreen instead `func gpuRect() -> GPU_Rect {` and grab `target` internally. Just readability wise it looks funky to me to have `gpuRect(target: rawPointer)` everywhere. To me, it's not really clear what that means without reading the function's internals.

ephemer (approved):
LGTM! but I think we can clean up the readability somewhat still

PR #415, opened 2026-06-05
Font fallback for missing glyphs via system fonts, larger right nav button tap target

ephemer:
- When the bundled font is missing a glyph, fall back to system fonts (NotoSansSymbols, Roboto, DroidSans) for both measurement and rendering
- Split text into same-font runs and render each with `TTF_RenderUTF8_Blended` (simpler, faster rendering)
- Give the navigation bar right button a minimum tap target matching bar height

### DIFF
--- a/Sources/FontRenderer+singleLineSize.swift
+++ b/Sources/FontRenderer+singleLineSize.swift
@@ -104,17 +104,16 @@ extension FontRenderer {
                 }
             }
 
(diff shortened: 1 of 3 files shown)


ephemer:
@michaelknoch this change fixes some UI I'm working on, but will also allow us to remove the `@SafeForAndroid` Strings we use elsewhere and render actual musical symbols via this fallback

michaelknoch (approved):

PR #416, opened 2026-06-12
Fix video seeks landing on the nearest keyframe instead of the requested time

AKiranB:
**Type of change:** bugfix

Fixes an issue where video seeks would be landing on the nearest keyframe instead of the requested time. This change make seeking more exact

### DIFF
--- a/src/main/java/org/uikit/VideoJNI.kt
+++ b/src/main/java/org/uikit/VideoJNI.kt
@@ -70,7 +70,7 @@ class AVPlayer(sdlView: SDLActivity, asset: AVURLAsset) {
                 newPosition: Player.PositionInfo,
                 reason: Int
             ) {
-                if (reason == Player.DISCONTINUITY_REASON_SEEK) {
+                if (reason == Player.DISCONTINUITY_REASON_SEEK_ADJUSTMENT) {
                     isSeeking = false
                     val delta = (newPosition.positionMs - desiredSeekPosition).absoluteValue
                     if (delta > 50) {


src/main/java/org/uikit/VideoJNI.kt
```diff
@@ -133,11 +147,11 @@ class AVPlayer(sdlView: SDLActivity, asset: AVURLAsset) {
         mainHandler.post(pendingSeek!!)
     }
 
-    private fun doSeekSync(timeMs: Long) {
+    private fun doSeekSync(timeMs: Long, forceExact: Boolean = false) {
         if (isSeeking) return
 
         val deltaTotal = (timeMs - lastSeekedToTime).absoluteValue
-        val params = if (deltaTotal < 100) SeekParameters.EXACT else SeekParameters.CLOSEST_SYNC
+        val params = if (forceExact || deltaTotal < 100) SeekParameters.EXACT else SeekParameters.CLOSEST_SYNC
```
michaelknoch: I wonder, should we kick the delta and always seek exact? I think this had a huge performance impact back then

michaelknoch (approved):
lgtm, if it fixes our issue.


Exoplayer introduced a scrubbing mode some time ago. Just from their docs it's exactly what we need, and maybe it makes our code here easier. When I tried it last time, I ran into other funky issues, but maybe worth a try?

https://android-developers.googleblog.com/2025/08/media3-180-whats-new.html

src/main/java/org/uikit/VideoJNI.kt
michaelknoch: if we would always seek exact, `DISCONTINUITY_REASON_SEEK_ADJUSTMENT` would never happen, I think

AKiranB: I think always seeking exact has some relatively serious performance implications, though. Especially when scrubbing

AKiranB:
I'll take a look at scrubbing

PR #418, opened 2026-06-15
[UIFont] Add a fallback scale to UIScreen.main.scale if the screen is inactive

ephemer:
**Type of change:** Bug fix

Apps will crash if fonts are constructed while UIScreen is deinited (e.g. the app is in the background). This change provides a sane fallback as a default to prevent crashes (at the potential cost of correctness in the worst case).

### DIFF
--- a/Sources/UIFont.swift
+++ b/Sources/UIFont.swift
@@ -27,7 +27,7 @@ open class UIFont {
 
     public init?(name: String, size: CGFloat) {
         let name = name.lowercased()
-        let size = Int32(size * UIScreen.main.scale)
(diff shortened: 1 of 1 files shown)


ephemer:
@michaelknoch we see this crash when connecting MIDI devices while the app is in the background

michaelknoch (approved):
I remember we decided against this fallback years ago, but tbh I think it makes sense

ephemer:
@michaelknoch it'd be better not to have the fallback, because it's possible to enter an invalid UI state where some fonts are rendered at different sizes to others with it present. I think the real solution is to make UIScreen non-optional and use a different abstraction for "we can render or not". But that's a bigger refactor

PR #420, opened 2026-07-02
improve text rendering

michaelknoch:
**Type of change:** feature

- Fix rounding for close button
- Allow center alignment of multiline text
- Implement `lineBreakMode byTruncatingTail` 
- Implement attributed text to allow rendering multiple fonts in one label text 

(p.s. The headline is repeated 2x in the screenshot to demonstrate the line truncation)
Screenshot 2026-07-02 at 09 42 51

### DIFF
--- a/Sources/FontRenderer.swift
+++ b/Sources/FontRenderer.swift
@@ -111,11 +111,17 @@ public class FontRenderer {
         return false
     }
 
-    func render(_ text: String?, color: UIColor, wrapLength: Int = 0) -> CGImage? {
+    func render(_ text: String?, color: UIColor, wrapLength: Int = 0, alignment: NSTextAlignment = .left) -> CGImage? {
         guard let text = text else { return nil }
 
         if wrapLength <= 0 && needsFallback(text) {
             return renderWithFallback(text, color: color)
+        }
+
+        // SDL_ttf renders wrapped lines left-aligned only. For centered/right text we compose the
+        // lines ourselves so each line is aligned within the wrap width.
+        if wrapLength > 0 && alignment != .left {
+            return renderWrapped(wrapLines(of: text, wrapLength: wrapLength), color: color, wrapLength: wrapLength, alignment: alignment)
         }
 
         guard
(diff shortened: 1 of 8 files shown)


Sources/UIScreen+render.swift
```diff
@@ -87,7 +87,10 @@ extension UIScreen {
             let previousProgram = ShaderProgram.currentlyActive
             ShaderProgram.roundedRect.activate()
             ShaderProgram.roundedRect.setFill(rect: rect, cornerRadius: cornerRadius)
-            GPU_RectangleFilled(rawPointer, gpuRect(rect), color: color.sdlColor)
+            // Draw into a quad 1px larger on every side so the SDF's antialiased edge isn't
+            // clipped at the shape's cardinal points (most visible on perfect circles). The
+            // extra ring is effectively zero-alpha for the shape itself.
+            GPU_RectangleFilled(rawPointer, gpuRect(rect.insetBy(dx: -1, dy: -1)), color: color.sdlColor)
```
michaelknoch: This image is heavily zoomed in, thats why it's pixelated. This shows the shape without the fix, the left border is cut-off 
Image

michaelknoch: I don't know if this change is just fighting a symptom tbh

Sources/FontRenderer.swift
```diff
@@ -261,3 +351,136 @@ extension FontRenderer {
         return CGSize(width: wrapLength, height: combinedTextHeight + combinedLineSpacing)
     }
 }
+
+@MainActor
+public class NSAttributedString {
```
ephemer, line 358: hmm. are we sure `NSAttributedString` is `@MainActor` in its Foundation implementation? It's not impossible but I would be surprised

Sources/UIFont.swift
```diff
@@ -6,7 +6,7 @@ open class UIFont {
     }
     public var pointSize: CGFloat
     public var lineHeight: CGFloat {
-        return CGFloat(renderer?.getLineHeight() ?? 0) / UIScreen.main.scale
+        return CGFloat(renderer?.getLineHeight() ?? 0) / (UIScreen.main?.scale ?? 2)
```
ephemer: This is unfortunately still incorrect and a growing source of potential "incorrectness".

A better solution here would be to have a `@MainActor static internal variable lastKnownScreenScale: CGFloat` which itself defaults to `2.0` but is set whenever `UIScreen.main` is initialized. Then we should reference that cached value internally whenever we don't have a screen. I know it's not directly related to this PR but it would be amazing if this could happen sooner rather than later, so I would also welcome that change in this PR 🙏🏼

Sources/UIScreen+render.swift
ephemer: It seems like it is fighting a symptom – it sounds like the antialising shader is antialiasing "outward", which isn't unreasonable but could cause this. I wonder if it's easy enough to adjust that shader rather than fudging pixels in seemingly unrelated places?

ephemer (approved):
Looks good! Left some minor comments that aren't really blocking, but would be good to see improvements in sooner rather than later 🙏🏼 Thanks for this @michaelknoch !

Sources/FontRenderer.swift
michaelknoch: Good point, you're right, Foundation's NSAttributedString isn't @MainActor; it's an immutable, Sendable type. I've dropped the annotation to match.

PR #421, opened 2026-07-08
fix button label truncation

michaelknoch:
**Type of change:** bugfix

**Issue:** a single-line `Button` title (or a `UILabel` laid out inside a button) was truncated with a trailing ellipsis even when its bounds were wide enough for the text. Expected: the full title shows, and the ellipsis only appears when the text genuinely doesn't fit.

**Cause:** the custom tail-truncation decided to truncate against a slightly different width than the one used to *size* the label, so a label sized to *just* fit read as overflowing by a pixel or two. Three small inconsistencies compounded it:

- the label was sized in one font weight (e.g. a caller reserving the bold/selected-state width) but rendered in another;
- `truncateTextIfNeeded` measured with `TTF_SizeUTF8`, a different ruler than the `size(_:)` used to size the label;
- the points→pixels conversion floored instead of rounded, dropping a sub-pixel on fractional screen scales (e.g. ~2.63x).

UIKit on iOS never shows this because `UIButton`/`UILabel` size and truncate through one consistent engine. This PR brings the cross-platform implementation in line: `Button` now sizes its title label to its current text/font each layout pass (as `UIButton` does), `truncateTextIfNeeded` measures with the same `size(_:)` ruler used for sizing, and the pixel conversion rounds instead of flooring. Net: labels show in full when they fit, and the ellipsis appears only on genuine overflow.

### DIFF
--- a/Sources/FontRenderer.swift
+++ b/Sources/FontRenderer.swift
@@ -136,19 +136,20 @@ public class FontRenderer {
 
     /// Returns `text` truncated with a trailing ellipsis so it fits within `wrapLength` (pixels).
     /// Returns `text` unchanged when it already fits (or `wrapLength <= 0`).
-    func truncateTextIfNeeded(_ text: String, wrapLength: Int) -> String {
+    func truncateTextIfNeeded(_ text: String, wrapLength: CGFloat) -> String {
         guard wrapLength > 0 else { return text }
 
-        // Measure with `size(_:)`, the same ruler that sized the label. `TTF_SizeUTF8` measures
-        // subtly differently, so a label sized to just fit could still be truncated here.
-        if Int(size(text).width) <= wrapLength { return text }
+        // Measure with `size(_:)`, the same ruler that sized the label; `TTF_SizeUTF8` measures
+        // subtly differently, so a label sized to just fit could still be truncated here. `wrapLength`
+        // is the available width in whole device pixels (the caller rounds up), so a fit compares cleanly.
+        if CGFloat(size(text).width) <= wrapLength { return text }
 
         let ellipsis = "…"
         var characters = Array(text)
         while !characters.isEmpty {
             characters.removeLast()
             let candidate = String(characters) + ellipsis
-            if Int(size(candidate).width) <= wrapLength { return candidate }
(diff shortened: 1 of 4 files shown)


Sources/UILabel.swift
```diff
@@ -79,7 +79,10 @@ open class UILabel: UIView {
         // UIKit's default behaviour instead of overflowing/clipping.
         var textToRender = text
         if numberOfLines == 1, lineBreakMode == .byTruncatingTail, let text = text, bounds.width > 0, let renderer = font.fontRenderer {
-            textToRender = renderer.truncateTextIfNeeded(text, wrapLength: Int(bounds.width * (UIScreen.lastKnownScreenScale ?? 2)))
+            // Round (not floor) points→pixels: width is pixels ÷ scale, so on a fractional scale
+            // `width * scale` can land a hair low and flooring would clip text that fits.
+            let availableWidthInPixels = Int((bounds.width * (UIScreen.lastKnownScreenScale ?? 2)).rounded())
```
michaelknoch: do we need rounding at all? rounding makes sense for rendering, but for layouting?

Sources/FontRenderer.swift
```diff
@@ -139,18 +139,16 @@ public class FontRenderer {
     func truncateTextIfNeeded(_ text: String, wrapLength: Int) -> String {
         guard wrapLength > 0 else { return text }
 
-        var width: Int32 = 0
-        var height: Int32 = 0
-        TTF_SizeUTF8(rawPointer, text, &width, &height)
-        if Int(width) <= wrapLength { return text }
+        // Measure with `size(_:)`, the same ruler that sized the label. `TTF_SizeUTF8` measures
+        // subtly differently, so a label sized to just fit could still be truncated here.
+        if Int(size(text).width) <= wrapLength { return text }
```
ephemer: if we're converting Double to Int, we need to explicitly use `.rounded()`, otherwise it'll be floored again implicitly, which is the bug we're trying to fix?

ephemer (approved):
Looks good, but let's make `wrapLength` a Double, then we don't need rounding at all here 👌🏼

PR #423, opened 2026-07-08
 Render CALayer drop shadows (+ UIBezierPath)

michaelknoch:
**Type of change:** feature

**Current:** the SDL renderer never shows a layer's shadow. It only drew one when `shadowPath` was set — and even then as a plain, unblurred fill with a hardcoded corner radius — and ignored `shadowColor` / `shadowOffset` / `shadowRadius` / `shadowOpacity`, which Core Animation uses to draw shadows automatically on iOS. `shadowPath` couldn't be set cross-platform anyway (`UIBezierPath` didn't exist; `CGPath` was just `typealias CGPath = CGRect`), so nothing rendered.

**Expected:** shadows render from the standard UIKit API and match iOS on both platforms.

## Fix

- A new SDF shadow shader draws a feathered fill behind the layer, honouring `shadowColor` / `shadowOffset` / `shadowRadius` / `shadowOpacity`, following `shadowPath` when set (otherwise the layer's rounded bounds) and clipped by `masksToBounds` — as on iOS. The blur is a smoothstep approximation of a Gaussian.
- `UIBezierPath` with the shape initializers (`rect` / `roundedRect` / `ovalIn`) plus a rounded-rect-capable `CGPath`, so `shadowPath` is `CGPath?` cross-platform. Freeform paths (move/addLine/curves) aren't modelled.

### DIFF
--- a/Sources/CALayer+SDL.swift
+++ b/Sources/CALayer+SDL.swift
@@ -73,6 +73,21 @@ extension CALayer {
             }
         }
 
+        // Core Animation draws a layer's shadow automatically on iOS; the SDL renderer doesn't, so we
+        // render a feathered fill *behind* the layer's own background, honouring shadowColor / Offset /
+        // Radius / Opacity and shadowPath. The shadow follows `shadowPath` if set (its own rect and
+        // corner radius, given in the layer's coordinate space), otherwise the layer's rounded bounds.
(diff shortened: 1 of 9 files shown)


ephemer (approved):
Nice!

PR #424, opened 2026-07-18
Add multitouch support

michaelknoch:
Adds two-finger touch handling to the SDL UIKit polyfill so views can pinch-to-zoom on Android/Mac:

- Build one shared multi-finger `UIEvent` from SDL finger events and deliver it through `sendEvent` (each event marks which finger changed).
- Add a real `UIPinchGestureRecognizer` driven by `event.allTouches`, a pan recognizer that locks to a single finger, and simultaneous pan+pinch coordination.

Also fixes `UIView.convert(_:to:)` to apply the **full** ancestor transform chain (lift to root, then descend to the target) instead of a single level, so points convert correctly through a scaled ancestor (e.g. a zoomed view) — replacing the old one-level hack.

### DIFF
--- a/Sources/UIPanGestureRecognizer.swift
+++ b/Sources/UIPanGestureRecognizer.swift
@@ -57,7 +57,8 @@ open class UIPanGestureRecognizer: UIGestureRecognizer {
 
     open override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent) {
         super.touchesBegan(touches, with: event)
-        guard let trackedTouch = touches.first else { return }
+        // A pan follows a single finger; ignore additional fingers (e.g. the second finger of a pinch).
+        guard trackedTouch == nil, let trackedTouch = touches.first else { return }
         self.trackedTouch = trackedTouch
         initialTouchPoint = trackedTouch.location(in: nil)
         touchesMovedTimestamp = trackedTouch.timestamp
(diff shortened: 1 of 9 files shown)


Sources/UIPanGestureRecognizer.swift
```diff
@@ -68,11 +69,10 @@ open class UIPanGestureRecognizer: UIGestureRecognizer {
         super.touchesMoved(touches, with: event)
         guard
             let trackedTouch = trackedTouch,
-            touches.first == trackedTouch,
+            touches.contains(trackedTouch),
             let initialTouchPoint = initialTouchPoint
         else {
-            state = .failed
```
ephemer, line 74: when does `state` ever get set to `.failed` then? or doesn't it now? does it matter?

ephemer (approved):
LGTM! Just unsure about the .failed lifecycle question

Sources/UIPanGestureRecognizer.swift
michaelknoch: yeah this is intentional. on real ios the pan doesn't fail when a second finger moves — it keeps going (it tracks the average position of all fingers, the "centroid"). the only case ios uses `.failed` for is "this can't be a pan" (wrong touch count, or `require(toFail:)` letting another recognizer win) — and `require(toFail:)` is an empty stub in our code, so nothing here ever depends on a pan failing. We might want to come back to this when we properly implement require(toFail:)

the old else-branch was the wrong bit: a second finger moving flipped the pan to `.possible` mid-gesture, but tracking state stayed set, so it could never get back to `.changed`, the pan just died the moment you added the pinch finger. ignoring touches that aren't ours keeps it alive.

michaelknoch: long story short: I think it doesn't matter

PR #425, opened 2026-07-20
Gradient layer gpu and mask support

AKiranB:
Fixes # https://app.asana.com/1/2883569378809/project/1206248656749420/task/1216650934897706

**Type of change:** Update

We want to rasterize the gradient via the GPU, rather than the nested pixel loop the status quo was doing.
The general idea is to rasterize the gradient once into a cached texture (until the panel itself changes) and then copy/blitting this for each frame.

Screenshot 2026-07-23 at 16 06 20

### DIFF
--- a/Sources/CAGradientLayer.swift
+++ b/Sources/CAGradientLayer.swift
@@ -68,19 +68,31 @@ open class CAGradientLayer: CALayer {
 
         guard let target = GPU_GetTarget(image.rawPointer) else { return }
 
-        // Disable the target's camera (inverted on image targets) so a plain ortho + identity maps our
-        // rect 1:1 onto the texture; otherwise the shader's camera × projection × modelview draws off-target.
+        // This runs mid-parent-render-pass, so save every piece of shared GL state we touch and restore it
+        // via `defer` — that way any early return added below can't leak state into the outer pass.
         let savedClippingRect = UIScreen.main?.clippingRect
         UIScreen.main?.clippingRect = nil
         GPU_FlushBlitBuffer()
-
-        GPU_EnableCamera(target, false)
-        GPU_MatrixMode(GPU_PROJECTION); GPU_PushMatrix(); GPU_LoadIdentity()
-        GPU_Ortho(0, Float(pixelWidth), Float(pixelHeight), 0, -1, 1)
-        GPU_MatrixMode(GPU_MODELVIEW); GPU_PushMatrix(); GPU_LoadIdentity()
-
-        GPU_SetViewport(target, GPU_Rect(x: 0, y: 0, w: Float(pixelWidth), h: Float(pixelHeight)))
+        GPU_MatrixMode(GPU_PROJECTION); GPU_PushMatrix()
+        GPU_MatrixMode(GPU_MODELVIEW); GPU_PushMatrix()
         GPU_SetShapeBlending(false)
+        defer {
+            ShaderProgram.deactivateAll()
+            GPU_FlushBlitBuffer()
+            GPU_MatrixMode(GPU_MODELVIEW); GPU_PopMatrix()
+            GPU_MatrixMode(GPU_PROJECTION); GPU_PopMatrix()
+            GPU_MatrixMode(GPU_MODELVIEW) // leave mode as the outer render pass expects it
+            GPU_SetShapeBlending(true)
+            UIScreen.main?.clippingRect = savedClippingRect
+        }
+
+        // Disable the target's camera (inverted on image targets) so a plain ortho + identity maps our
+        // rect 1:1 onto the texture; otherwise the shader's camera × projection × modelview draws off-target.
+        GPU_EnableCamera(target, false)
+        GPU_MatrixMode(GPU_PROJECTION); GPU_LoadIdentity()
+        GPU_Ortho(0, Float(pixelWidth), Float(pixelHeight), 0, -1, 1)
+        GPU_MatrixMode(GPU_MODELVIEW); GPU_LoadIdentity()
+        GPU_SetViewport(target, GPU_Rect(x: 0, y: 0, w: Float(pixelWidth), h: Float(pixelHeight)))
         GPU_Clear(target)
 
         ShaderProgram.gradient.activate()
@@ -94,15 +106,6 @@ open class CAGradientLayer: CALayer {
                 : colors.indices.map { Float($0) / Float(max(colors.count - 1, 1)) }
         )
         GPU_RectangleFilled(target, GPU_Rect(x: 0, y: 0, w: Float(pixelWidth), h: Float(pixelHeight)), color: UIColor.white.sdlColor)
-
-        ShaderProgram.deactivateAll()
-        GPU_FlushBlitBuffer()
-
-        GPU_MatrixMode(GPU_MODELVIEW); GPU_PopMatrix()
-        GPU_MatrixMode(GPU_PROJECTION); GPU_PopMatrix()
-        GPU_MatrixMode(GPU_MODELVIEW)
-        GPU_SetShapeBlending(true)
-        UIScreen.main?.clippingRect = savedClippingRect
 
         contents = image
     }
--- a/Sources/CAGradientLayer.swift
+++ b/Sources/CAGradientLayer.swift
@@ -1,3 +1,4 @@
+internal import SDL
 private import SDL_gpu
 
 open class CAGradientLayer: CALayer {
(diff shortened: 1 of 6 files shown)


Sources/CAGradientLayer.swift
```diff
@@ -46,127 +45,127 @@ open class CAGradientLayer: CALayer {
...
-        }
+extension ShaderProgram {
+    private static var _gradient: GradientShaderProgram?
+    static var gradient: GradientShaderProgram {
+        if let existing = _gradient { return existing }
+        let program = try! GradientShaderProgram()
+        _gradient = program
+        return program
+    }
+    static func invalidateGradient() { _gradient = nil }
```
michaelknoch, line 95: `invalidateGradient()` is defined but never called. `UIScreen.deinit` invalidates roundedRect/shadow/mask and calls `FragmentShader.invalidateAll()` (which resets `_gradient` the *fragment shader*), but the linked `GradientShaderProgram` cached here is never invalidated. After the GLContext is recreated (routine on Android) this holds a dead `programRef` → broken/garbage gradients. Add `ShaderProgram.invalidateGradient()` to `UIScreen.deinit` alongside the others.

Sources/MaskingShaders.swift
```diff
@@ -165,4 +179,50 @@ extension FragmentShader {
...
+            float t = len2 > 0.00001 ? dot(p - startPoint, dir) / len2 : 0.0;
+            t = clamp(t, 0.0, 1.0);
+
+            vec4 color = colors[0];
+            if (t <= locations[0]) {
+                color = colors[0];
+            } else if (t >= locations[colorCount - 1]) {
+                color = colors[colorCount - 1];
+            } else {
+                for (int i = 0; i < colorCount - 1; i++) {
```
michaelknoch: GLES2 portability risk. Under GLSL ES 2.0, uniform-array indices must be *constant-index-expressions*: a loop index qualifies (`colors[i]` is fine), but `colorCount - 1` is uniform-derived and does **not** — `locations[colorCount - 1]` / `colors[colorCount - 1]` may fail to compile on some Android drivers, which then hits the `try!` and crashes. The non-constant loop bound `i < colorCount - 1` is the same class of issue. Please test on a real GLES2 device, or restructure to read the last stop via the loop instead of a uniform-indexed access.

Sources/CALayer+SDL.swift
```diff
@@ -67,9 +67,15 @@ extension CALayer {
             renderer.clippingRect =
                 renderer.clippingRect?.intersection(maskAbsoluteFrame) ?? maskAbsoluteFrame
 
+            // The mask isn't in the sublayer tree, so render it here to populate its `contents`.
+            if mask.contentsScale != contentsScale { mask.contentsScale = contentsScale }
+            if mask.needsDisplay() { mask.display(); mask._needsDisplay = false }
+
             if let maskContents = mask.contents {
```
michaelknoch, line 77: Nit: stray whitespace-only line directly below this.

Sources/MaskingShaders.swift
```diff
@@ -84,7 +97,8 @@ extension FragmentShader {
             );
 
             vec4 maskColour = \(texture)(maskTexture, maskCoordinate);
-            \(fragColor) = vec4(originalColour.rgb, originalColour.a * maskColour.a);
+            vec4 base = \(texture)(tex, texCoord) * originalColour;
```
michaelknoch: This is the biggest change in the PR and it affects **every masked layer in the app**, not just the new gradient panel — `ShaderProgram.mask` is a singleton used by all masks.

Two things to check:

1. **Behavioural change for textured masks.** Previously the mask shader output `originalColour.rgb` (the per-vertex colour), so a *blitted* masked layer rendered as flat white/tint, ignoring its own texture. Now it multiplies by `texture(tex, texCoord)`, so the image shows through. That's a genuine fix, but any existing masked view needs a visual re-check for regressions.

2. **Undefined `tex`/`texCoord` for non-textured draws under the mask.** SDL_gpu routes untextured shape draws (`GPU_RectangleFilled`/`GPU_Rectangle`) through the active shader too. In `CALayer.sdlRender`, a plain (cornerRadius < 1) `backgroundColor` fill and a thin border stay on the mask shader (only the rounded-rect/shadow paths swap shaders and restore). For those draws there's no bound texture at unit 0 and `gpu_TexCoord` is undefined → `texture(tex, texCoord)` samples garbage/stale data, which can tint or zero-out (`base.a`) the fill. Please verify a masked layer with a non-rounded background/border still renders, or guard the sampler so shape draws fall back to `originalColour`.

Sources/MaskingShaders.swift
```diff
@@ -61,20 +63,31 @@ extension FragmentShader {
...
     private static let maskColourWithImageSource = """
         \(`in`) vec4 originalColour;
         \(`in`) vec2 absolutePixelPos;
+        \(`in`) vec2 texCoord;
 
         \(fragColorDefinition)
 
         uniform vec4 maskFrame;
         uniform sampler2D maskTexture;
+        uniform sampler2D tex; // the masked layer's own contents (blit image, texture unit 0)
```
michaelknoch: `tex` is never assigned a texture unit and relies on the GLSL default (unit 0) coinciding with the unit SDL_gpu blits into, while `maskTexture` is explicitly bound to unit 1 (`UniformVariable.set(CGImage)`). Worth making this explicit (or at least asserting the location), since it's a silent-breakage point if SDL_gpu's blit unit ever changes.

Sources/CALayer+SDL.swift
```diff
@@ -67,9 +67,15 @@ extension CALayer {
...
+             
+                let maskShaderFrame = maskFrame.offsetBy(deltaFromAnchorPointToOrigin)
```
michaelknoch, line 81: Mask frame semantics changed from `mask.bounds` to `maskFrame.offsetBy(deltaFromAnchorPointToOrigin)`. This looks like a real alignment fix (the masked content is drawn starting at `deltaFromAnchorPointToOrigin`, not at 0,0), but it changes where **all** existing masks sample. Please eyeball existing mask users to confirm they still line up, especially any with a non-default `anchorPoint` or a mask `frame.origin != 0`.

Sources/CALayer+SDL.swift
```diff
@@ -67,9 +67,15 @@ extension CALayer {
             renderer.clippingRect =
                 renderer.clippingRect?.intersection(maskAbsoluteFrame) ?? maskAbsoluteFrame
 
+            // The mask isn't in the sublayer tree, so render it here to populate its `contents`.
+            if mask.contentsScale != contentsScale { mask.contentsScale = contentsScale }
+            if mask.needsDisplay() { mask.display(); mask._needsDisplay = false }
```
michaelknoch, line 75: The mask is now driven purely through `display()`/`contents` here — it's never `sdlRender`'d, so a mask that relies on its own `backgroundColor`, border, or sublayers (rather than `contents`) won't be honoured. That's fine for `CAGradientLayer`/image masks, but worth a one-line note so nobody expects full-subtree masking.

Also: `mask.display()` now runs a full GPU render-to-target (matrix push/pop, viewport, clippingRect + shape-blending toggles) *in the middle of the parent's render pass*. It does restore state today, but this reentrancy is fragile — any future early-return in `display()` that skips the restore would corrupt the outer pass. Consider a `defer`-based restore in `CAGradientLayer.display()`.

Sources/CAGradientLayer.swift
```diff
@@ -46,127 +45,127 @@ open class CAGradientLayer: CALayer {
...
+
+        ShaderProgram.gradient.activate()
+        ShaderProgram.gradient.set(
+            rect: CGRect(x: 0, y: 0, width: pixelWidth, height: pixelHeight),
+            startPoint: startPoint,
+            endPoint: endPoint,
+            colors: _colors,
+            locations: locations.count >= colors.count
+                ? locations
+                : colors.indices.map { Float($0) / Float(max(colors.count - 1, 1)) }
```
michaelknoch: Minor: this `locations.count >= colors.count ? ... : recompute` is dead — `display()` already normalises `locations` to `colors.count` earlier (the `if locations.count < colors.count` block), so by here the ternary is always the first branch. Can just pass `locations`.

Sources/MaskingShaders.swift
AKiranB: Good point, tbf

Sources/CAGradientLayer.swift
AKiranB: Also good catch

michaelknoch (approved):
LGTM code wise! Did you test if the shader recovers when putting the app in background and foreground again?

Sources/ShaderProgram+mask.swift
```diff
@@ -8,30 +8,45 @@
...
+    fileprivate init(fragmentShader: FragmentShader) throws {
+        try super.init(vertexShader: .common, fragmentShader: fragmentShader)
         maskFrame = UniformVariable("maskFrame", in: programRef)
         maskTexture = UniformVariable("maskTexture", in: programRef)
+        texLocation = GPU_GetUniformLocation(programRef, "tex")
     }
 
     func set(maskImage: CGImage, frame: CGRect) {
-        maskFrame.set(frame)
+        maskFrame.set(frame)               // maskTexture is bound to texture unit 1 (see UniformVariable.set)
```
michaelknoch, line 46: wild whitespace

AKiranB: 😆  Might have been me by accident

Sources/CAGradientLayer.swift
```diff
@@ -46,127 +45,128 @@ open class CAGradientLayer: CALayer {
...
-        ) else {
-            return
+        guard let target = GPU_GetTarget(image.rawPointer) else { return }
+
+        // This runs mid-parent-render-pass, so save every piece of shared GL state we touch and restore it
+        // via `defer` — that way any early return added below can't leak state into the outer pass.
+        let savedClippingRect = UIScreen.main?.clippingRect
+        UIScreen.main?.clippingRect = nil
+        GPU_FlushBlitBuffer()
+        GPU_MatrixMode(GPU_PROJECTION); GPU_PushMatrix()
```
ephemer: this dance around MatrixMode and pushing and popping matrices looks surprising to me. do we really need it? this is the only part of the PR that i feel uncomfortable about now that all the other improvements have landed.

ephemer (approved):
Generally it looks good to me, I'm just unsure about one part which I left a comment on

PR #429, opened 2026-07-30
Fix Android SIGSEGV: don't free CGImage textures against a destroyed GL context

tanarie:
**Type of change:** bug fix

## Motivation

Fixes the ongoing Android crash:
https://app.bugsnag.com/flowkey/mobile-app/errors/69644a5e4ef4ae6af7fd0f05

`SIGSEGV` in `renderer_GL_common.inl:3923`, inside `GPU_FreeImage`. Android only.
Last 30 days: 108 events, 106 users; 198 events / 192 users all time. First seen
2026-01-12, still happening on 2.113.2. It correlates with the app going to the
background and coming back.

### Root cause

On Android, backgrounding tears down the `UIScreen` and its GL context
(`UIScreen.main = nil` → `UIScreen.deinit` → `GPU_Quit()`); resuming builds a
**new** `UIScreen`/context. A `CGImage` created under the **old** context can
then be freed *after* the new context exists, and `CGImage.deinit`'s guard only
checked `UIScreen.main != nil` — which passes, because a (new) screen exists. So
`GPU_FreeImage` ran against an image whose `context_target` is freed memory →
crash. The guard asked "does *a* screen exist?", not "is the context that owns
*this* image still alive?".

Why it reliably lands after the resume: `deinit` frees via `Task { @MainActor }`,
and Android pauses our render loop while backgrounded, so jobs queued during
teardown only drain on the first frame after resume — the stack trace shows the
free inside `nativeProcessEventsAndRender`.

Why sdl-gpu's own guard doesn't help: it compares
`image->renderer == GPU_GetCurrentRenderer()`, but the renderer is `SDL_malloc`'d
and `SDL_free`'d (`renderer_GLES_2.c:32`/`:62`), so the `GPU_Init` after
`GPU_Quit` usually gets the same address back and the comparison passes
spuriously.

### Fix

Tag each `CGImage` with the `UIScreen.contextGeneration` it was created under,
and only call `GPU_FreeImage` when that generation is still the current one —
in both `deinit` and `reloadFromSourceData()`. The counter is bumped whenever a
GL context is created *or destroyed*, so "generation matches" means exactly
"the context that owns this image is still live", without depending on
`UIScreen.main` being nil for the whole teardown→reinit window.

Skipping the free doesn't leak texture memory (that goes with the context), but
it does leak the small `GPU_Image`/`GPU_IMAGE_DATA` structs — `FreeImage`'s two
`SDL_free` calls sit outside the GL branch. ~100 bytes per live image per
background cycle. Freeing them properly needs a registry of live `CGImage`s
torn down before `GPU_Quit()`; follow-up, not a blocker.

## Notes / testing
- Couldn't reproduce locally (POCO M3 / Adreno). Adreno keeps the GL context on
  background, so it rarely triggers; the crash mostly hits Samsung/Mali devices.
- Not verified locally — please review carefully, and confirm via the Bugsnag
  count dropping after release.
- Follow-up: bump submodules UIKit-SDL → NativePlayer → mobile-app once merged.

- [ ] There are no dependencies on other PRs

### DIFF
--- a/Sources/CGImage.swift
+++ b/Sources/CGImage.swift
@@ -119,8 +119,8 @@ public class CGImage {
 
         // Free the old GPU_Image before replacing it (this may be our last chance),
         // but only if it still belongs to the live context. If its context was already
-        // destroyed (stale generation) the texture is gone with it, and freeing it here
-        // would crash against a dangling context (see `deinit`).
+        // destroyed (stale generation), freeing it here would make GL calls against a
+        // dangling context and crash (see `deinit`).
         if contextGeneration == UIScreen.contextGeneration {
             GPU_FreeImage(rawPointer)
         }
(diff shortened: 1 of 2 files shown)


michaelknoch (approved):

michaelknoch:
great fix, thanks @tanarie

PR #430, opened 2026-08-04
Fix mask clip rect not scaling

AKiranB:
Fixes # 

Before
Screenshot 2026-08-04 at 18 31 02
After
Screenshot 2026-08-04 at 18 27 52

Sticky panel was clipped on Android tablets because the horizontal sheet scales up but he mask clip rect wasn't scaled to match. Now runs the mask frame through the same modelViewTransform as the layer's content.


michaelknoch (approved):
LGTM!  Fixes the bug on my test devices
thanks @AKiranB

ephemer (approved):

PR #431, opened 2026-08-06
implement CALayer+ContentsFilter

michaelknoch:
**Type of change:** feature

Contents drawn smaller than they were rendered are sampled bilinearly, with no way to change it. Minifying by much more than 1:1 undersamples — thin lines break up and crawl as the scale changes.

This adds Apple's `CALayer.minificationFilter`, so a layer can ask for `.trilinear`: mipmaps, blended between levels. Costs ~50% more texture memory (measured on Mali-G710), so `.linear` stays the default.

The caller is the FKDL sheet renderer in NativePlayer, which rasters tiles once for the sharpest zoom the sheet can reach and then only ever scales them down — minification is its normal state, not an edge case.

## Notes

- Filtering lives on the texture, not the layer, so an image shared by two layers takes the filter of whichever set `contents` last.
- Mipmaps can't be un-generated: after `.trilinear`, `.linear` samples them with nearest level selection rather than reverting to plain bilinear.
- `.nearest` and `magnificationFilter` are deliberately absent — SDL_gpu sets the minification and magnification filters from a single value, so neither would behave as it does on iOS. Failing to compile beats differing quietly.
- `replacePixels` does not regenerate mipmaps. Nothing pairs it with `.trilinear` today; two lines when something does.

### DIFF
--- a/Sources/UIScreen.swift
+++ b/Sources/UIScreen.swift
@@ -44,10 +44,26 @@ public final class UIScreen {
     }
     nonisolated public let scale: CGFloat
 
+    /// The GPU's `GL_MAX_TEXTURE_SIZE`. Callers rastering their own images have to respect it themselves:
+    /// sdl-gpu never validates a size, so `GPU_CreateImage` hands back a non-nil but incomplete texture
+    /// above the limit, which renders as garbage with no error anywhere.
+    nonisolated public let maxTextureSize: CGFloat
+
     private init(renderTarget: UnsafeMutablePointer<GPU_Target>!, bounds: CGRect, scale: CGFloat) {
         self.rawPointer = renderTarget
         self.bounds = bounds
         self.scale = scale
+        self.maxTextureSize = Self.getMaxTextureSize()
+    }
+
+    private static func getMaxTextureSize() -> CGFloat {
+        let GL_MAX_TEXTURE_SIZE: UInt32 = 0x0D33
+        typealias GetIntegerv = @convention(c) (UInt32, UnsafeMutablePointer<Int32>) -> Void
+        var value: Int32 = 0
+        unsafeBitCast(
+            SDL_GL_GetProcAddress("glGetIntegerv")!, to: GetIntegerv.self
+        )(GL_MAX_TEXTURE_SIZE, &value)
+        return CGFloat(value)
     }
 
     convenience init() {
--- a/Sources/UIScreen.swift
+++ b/Sources/UIScreen.swift
@@ -44,27 +44,16 @@ public final class UIScreen {
     }
     nonisolated public let scale: CGFloat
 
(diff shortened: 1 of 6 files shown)


Sources/UIScreen.swift
```diff
@@ -44,10 +44,27 @@ public final class UIScreen {
     }
     nonisolated public let scale: CGFloat
 
+    /// The GPU's `GL_MAX_TEXTURE_SIZE`. Callers rastering their own images have to respect it themselves:
+    /// sdl-gpu never validates a size, so `GPU_CreateImage` hands back a non-nil but incomplete texture
+    /// above the limit, which renders as garbage with no error anywhere.
+    nonisolated public let maxTextureSize: CGFloat
```
michaelknoch, line 50: this is actually unrelated to the ContentsFilter, but also used in the player PR. This is no official apple API 😢

Sources/UIScreen.swift
```diff
@@ -44,10 +44,27 @@ public final class UIScreen {
...
+        // No render target (the test `dummyScreen`) means no GL context to query; 0 signals "no GPU".
+        self.maxTextureSize = renderTarget != nil ? Self.getMaxTextureSize() : 0
+    }
+
+    private static func getMaxTextureSize() -> CGFloat {
+        let GL_MAX_TEXTURE_SIZE: UInt32 = 0x0D33
+        typealias GetIntegerv = @convention(c) (UInt32, UnsafeMutablePointer<Int32>) -> Void
+        var value: Int32 = 0
+        unsafeBitCast(
+            SDL_GL_GetProcAddress("glGetIntegerv")!, to: GetIntegerv.self
```
ephemer: this looks a bit funky. is there really no cleaner way of doing this?

ephemer (approved):
Looks good, i'm just unsure about the very low level "get gl function" call – does SDL-gpu not expose this?

Sources/UIScreen.swift
```diff
@@ -44,10 +44,16 @@ public final class UIScreen {
...
+    /// The GPU's `GL_MAX_TEXTURE_SIZE`, or 0 without a GPU context. Callers rastering their own images have
+    /// to respect it themselves: sdl-gpu never validates a size, so `GPU_CreateImage` hands back a non-nil
+    /// but incomplete texture above the limit, which renders as garbage with no error anywhere.
+    nonisolated public let maxTextureSize: CGFloat
+
     private init(renderTarget: UnsafeMutablePointer<GPU_Target>!, bounds: CGRect, scale: CGFloat) {
         self.rawPointer = renderTarget
         self.bounds = bounds
         self.scale = scale
+        self.maxTextureSize = CGFloat(GPU_GetMaxTextureSize())
```
michaelknoch, line 56: way cleaner @ephemer :)

